PR #4, opened 2023-06-07
uyuniadm migrate command for podman

mbussolotto:


uyuniadm/cmd/migrate/utils.go
```diff
@@ -0,0 +1,83 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+func getSshAuthSocket() string {
+	path := os.Getenv("SSH_AUTH_SOCK")
+	if len(path) == 0 {
+		log.Fatal("SSH_AUTH_SOCK is not defined, start an ssh agent and try again")
```
mbussolotto, line 13: are we sure about this? The reason why we're using a SSH Agent it's because it's the only way we found for running rsync passwordless and no-interactive, and the SSH Agent requires SSH_AUTH_SOCK...but if a user finds another way for doing that, why should we force him to create an SSH Agent? What do you think?

uyuni-tools.spec
```diff
@@ -0,0 +1,88 @@
...
+go build -o ./bin ./uyuniadm/ 
+go build -o ./bin ./uyunictl/ 
+
+%install
+install -m 0755 -vd                     %{buildroot}%{_bindir}
+install -m 0755 -vp ./bin/* %{buildroot}%{_bindir}/
+mkdir -p %{buildroot}%{_sysconfdir}/%{project}/
+install -m 0755 -vp ./uyuniadm/options.json  %{buildroot}%{_sysconfdir}/%{project}/options.json
+mkdir -p %{buildroot}%{_sharedstatedir}/%{project}/
+install -m 0755 -vp ./uyuniadm/setup-migration-container.sh %{buildroot}%{_sharedstatedir}/%{project}/setup-migration-container.sh
```
mbussolotto, line 69: Since `generateMigrationScript` generates the script, I think we can remove it.

rjmateus (approved):
Merged it and then we can work on top of it
Code is not yet release, so no big impact

uyuniadm/cmd/migrate/utils.go
cbosdo: They will tell us in an issue 😉  For now I would fail early if possible

PR #12, opened 2023-08-09
fix migration: new container should have same hostname and ssh passwordless fix

mbussolotto:
- the agent for working requires also the ssh key, so it may be better to mount the entire `.ssh` folder
- improve help
- add  `-o StrictHostKeyChecking=no` to rsync, to prevent any interaction
- migrate container should have same fqdn

### DIFF
--- a/uyuniadm/cmd/migrate/migrate.go
+++ b/uyuniadm/cmd/migrate/migrate.go
@@ -22,6 +22,7 @@ func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
 This migration command assumes a few things:
   * the SSH configuration for the source server is complete, including user and
     all needed options to connect to the machine,
+  * the SSH configuration should stored in $HOME/.ssh, including the keys
   * an SSH agent is started and the key to use to connect to the server is added to it,
   * podman or kubectl is installed locally
   * if kubectl is installed, a working kubeconfig should be set to connect to the cluster to deploy to
--- a/uyuniadm/cmd/migrate/podman.go
+++ b/uyuniadm/cmd/migrate/podman.go
@@ -19,15 +19,15 @@ import (
 func migrateToPodman(globalFlags *types.GlobalFlags, flags *flagpole, cmd *cobra.Command, args []string) {
 	sshAuthSocket := getSshAuthSocket()
 
+	sourceFqdn := args[0]
+
 	// Find ssh config to mount it in the container
 	homedir, err := os.UserHomeDir()
 	if err != nil {
 		log.Fatal("Failed to find home directory to look for SSH config")
 	}
-	sshConfigPath := filepath.Join(homedir, ".ssh", "config")
-	sshKnownhostsPath := filepath.Join(homedir, ".ssh", "known_hosts")
-
-	scriptDir := generateMigrationScript(args[0], false)
+	sshPath := filepath.Join(homedir, ".ssh")
+	scriptDir := generateMigrationScript(sourceFqdn, false)
 	defer os.RemoveAll(scriptDir)
 
 	extraArgs := []string{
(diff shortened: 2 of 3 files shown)


mbussolotto:
@cbosdo just to warn you, so we can discuss about this when you'll be back

mbussolotto:
still issues, since we're not longer settings the `hostname`. I'm going to fix it

uyuniadm/cmd/migrate/podman.go
```diff
@@ -60,7 +56,10 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *flagpole, cmd *cobra
 
 	image := fmt.Sprintf("%s:%s", flags.Image, flags.ImageTag)
 
-	podman.GenerateSystemdService(tz, image, viper.GetStringSlice("podman.arg"), globalFlags.Verbose)
+	podmanMigrateArgs := viper.GetStringSlice("podman.arg")
+	podmanMigrateArgs = append(podmanMigrateArgs, "--hostname="+sourceFqdn)
```
cbosdo, line 60: Using this parameter is dangerous: it will work for `podman` but not for Kubernetes. Couldn't we just set the proper parameters in the config as the setup is doing?

uyuniadm/cmd/migrate/utils.go
```diff
@@ -27,7 +27,7 @@ func generateMigrationScript(sourceFqdn string, kubernetes bool) string {
 set -e
 for folder in {{range .Volumes}}{{.}} {{end}};
 do
-  rsync -e "ssh -A " --rsync-path='sudo rsync' -avz {{.SourceFqdn}}:$folder/ $folder;
+  rsync -e "ssh -A -o StrictHostKeyChecking=no" --rsync-path='sudo rsync' -avz {{.SourceFqdn}}:$folder/ $folder;
```
cbosdo, line 30: We should not add this option, but let the user add it to his `ssh_config`

uyuniadm/cmd/migrate/utils.go
```diff
@@ -38,6 +38,9 @@ ssh {{ .SourceFqdn }} timedatectl show -p Timezone >/var/lib/uyuni-tools/data
 echo 'server.no_ssl = 1' >> /etc/rhn/rhn.conf;
 sed 's/address=[^:]*:/address=uyuni:/' -i /etc/rhn/taskomatic.conf;
 sed 's/address=[^:]*:/address=uyuni:/' -i /etc/sysconfig/tomcat;
+{{ else }}
+sed 's/address=[^:]*:/address={{ .SourceFqdn }}:/' -i /etc/rhn/taskomatic.conf;
+sed 's/address=[^:]*:/address={{ .SourceFqdn }}:/' -i /etc/sysconfig/tomcat;
```
cbosdo, line 43: Isn't that alone enough? I would think with those we can get rid of `--hostname` parameter.

cbosdo (commented):
Smells like there are dangerous bits there

uyuniadm/cmd/migrate/migrate.go
```diff
@@ -22,6 +22,7 @@ func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
 This migration command assumes a few things:
   * the SSH configuration for the source server is complete, including user and
     all needed options to connect to the machine,
+  * the SSH configuration should stored in $HOME/.ssh, including the keys
```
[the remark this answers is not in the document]
mbussolotto: mmmm I can re-try it, but IIRC it didn't works, also the keys were required

uyuniadm/cmd/migrate/utils.go
mbussolotto: Tested it. It works but the ssh_config should have the fqdn as host value. I'll change it and fix the documentation

uyuniadm/cmd/migrate/utils.go
mbussolotto: same issue of https://github.com/uyuni-project/uyuni-tools/pull/12#discussion_r1300121708 . We can get rid of it once we find a way to set the podman hostname

PR #14, opened 2023-08-10
Change cobra commands flags from StringArray to StringSlice

mbussolotto:
Change cobra commands flags from StringArray to StringSlice, to avoid issue on Viper when argument is empty


rjmateus (approved):
LGTM. merge it :)

PR #15, opened 2023-08-10
Code refactoring 

rjmateus:
Refactoring code.
Each commit has a refactoring.

Code should be compiling but I didn't test if anything breaks.

I have the code refactoring separated commits.
Each command from the uyuniadm is processing the parameters in a different way. I updated the "migrate" command to use viper unmarshal. I think we should do the same for the install and map the viper properties for a internal structure and pass that to the internal methods intead of rely in the viper object. That would make our code decoupled of viper library.


uyuniadm/shared/utils/cmd_utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
+
+import "github.com/spf13/cobra"
+
+func SetPodmanStartFlag (cmd *cobra.Command){
+	cmd.Flags().StringSlice("podman-arg", []string{}, "Extra arguments to pass to podman")
+}
+
+func SetHemlStartFlag (cmd *cobra.Command){
```
cbosdo: `s/Heml/Helm/`

uyuniadm/shared/utils/cmd_utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
```
cbosdo, line 1: Looks like this file should go in another commit

uyuniadm/cmd/install/install.go
```diff
@@ -87,6 +87,10 @@ NOTE: for now installing on a remote cluster or podman is not supported!
 	installCmd.Flags().String("scc-user", "", "SUSE Customer Center username")
 	installCmd.Flags().String("scc-password", "", "SUSE Customer Center password")
 
+	cmd_utils.AddImageFlag(installCmd)
+	cmd_utils.AddPodmanInstallFlag(installCmd)
+	cmd_utils.AddHemlInstallFlag(installCmd)
```
cbosdo, line 80: Typo! and the functions should be moved from the first  commit to this one.

uyuniadm/cmd/migrate/migrate.go
```diff
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/uyuni-project/uyuni-tools/shared/types"
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
+	cmd_utils "github.com/uyuni-project/uyuni-tools/uyuniadm/shared/utils"
 )
 
 type flagpole struct {
```
cbosdo, line 12: Probably this could go away now.

cbosdo (commented):
Not sure what you want to achieve with the last commit. I'll take care of the first ones though as they look good.

PR #18, opened 2023-08-21
Fix podman network setup to detect disabled ipv6

cbosdo:
If IPv6 is disabled on the host, creating a podman network with IPv6 enabled will lead to the bridge interface failing to go up.

Try hard to detect if IPv6 is disabled on the host before enabling it.

To help updating the configuration after enabling IPv6 on the host, also remove any existing `uyuni` network that doesn't have IPv6 and recreate if needed.

### DIFF
--- a/shared/podman/podman.go
+++ b/shared/podman/podman.go
@@ -70,7 +70,7 @@ ExecStart=/usr/bin/podman run \
 	-v {{ $name }}:{{ $path }} \
 	{{- end }}
 	-e TZ=${TZ} \
-	--network {{ .Network }}\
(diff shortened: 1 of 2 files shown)


rjmateus:
Please log which IP versions Will be uses. User should also force one or the other

cbosdo:
> Please log which IP versions Will be uses. User should also force one or the other

I'll log if IPv6 is not enabled, and that's it. The podman network will mimic what is available on the host. The less we let people tamper with the network definition the safer for us.

mbussolotto (approved):
LGTM

PR #26, opened 2023-09-08
add missing metrics ports for prometheus

rjmateus:
Add missing ports for JMX export for prometheus:
https://documentation.suse.com/suma/4.3/en/suse-manager/administration/monitoring.html

### DIFF
--- a/shared/utils/pod.go
+++ b/shared/utils/pod.go
@@ -45,6 +45,7 @@ func newPortMap(name string, exposed int, port int) PortMap {
 	}
 }
 
+// The port names should be less than 15 characters long and lowercased for traefik to eat them
 var TCP_PORTS = []PortMap{
 	newPortMap("postgres", 5432, 5432),
 	newPortMap("salt-publish", 4505, 4505),
@@ -52,8 +53,9 @@ var TCP_PORTS = []PortMap{
 	newPortMap("cobbler", 25151, 25151),
 	newPortMap("tomcat-debug", 8000, 8080),
 	newPortMap("tasko-debug", 8001, 8081),
-	newPortMap("psql-metrics", 9187, 9187),
-	newPortMap("node-metrics", 9101, 9101),
+	newPortMap("psql-mtrx", 9187, 9187),
+	newPortMap("tasko-jmx-mtrx", 5556, 5556),
+	newPortMap("tomcat-jmx-mtrx", 5557, 5557),
 }
 
 var UDP_PORTS = []PortMap{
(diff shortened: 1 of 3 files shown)


cbosdo (changes requested):
Could you please also add them to `shared/utils/pod.go` in the `TCP_PORTS`? I should refactor the podman code to use this structure too, but that can be in a separate PR. Just mind that the name in that struct is used by Kubernetes and needs to be short... but I can't remember the limit

shared/utils/pod.go
```diff
@@ -54,6 +54,10 @@ var TCP_PORTS = []PortMap{
 	newPortMap("tasko-debug", 8001, 8081),
 	newPortMap("psql-metrics", 9187, 9187),
 	newPortMap("node-metrics", 9101, 9101),
+	newPortMap("client-metrics", 9100, 9100),
```
cbosdo, line 57: From the name it looks like`node-metrics` and `client-metrics` are the same things. IIRC I had to use port 9101 because 9100 is already used by traefik on k3s.

cbosdo (commented):
Just a doubt on the 9100 port

shared/utils/pod.go
rjmateus: I have adjusted the helm definition: https://github.com/uyuni-project/uyuni/commit/afa18797d98b5b02c25cc54442f442defebd7cff

witekest: As replied per email. We should consider dropping node exporter in favor of cAdvisor.

shared/utils/pod.go
```diff
@@ -53,7 +53,9 @@ var TCP_PORTS = []PortMap{
 	newPortMap("tomcat-debug", 8000, 8080),
 	newPortMap("tasko-debug", 8001, 8081),
 	newPortMap("psql-metrics", 9187, 9187),
-	newPortMap("node-metrics", 9101, 9101),
+	newPortMap("node-metrics", 9100, 9100),
+	newPortMap("taskomatic-JMX-metrics", 5556, 5556),
+	newPortMap("tomcat-JMX-metrics", 5557, 5557),
```
cbosdo, line 58: These names should be less than 15 characters long and lowercase:
```
Error: INSTALLATION FAILED: 2 errors occurred:
	* Service "traefik" is invalid: [spec.ports[6].targetPort: Invalid value: "taskomatic-jmx-metrics": must be no more than 15 characters, spec.ports[8].targetPort: Invalid value: "tomcat-jmx-metrics": must be no more than 15 characters]
	* Deployment.apps "traefik" is invalid: [spec.template.spec.containers[0].ports[7].name: Invalid value: "taskomatic-jmx-metrics": must be no more than 15 characters, spec.template.spec.containers[0].ports[10].name: Invalid value: "tomcat-jmx-metrics": must be no more than 15 characters]
```

cbosdo (commented):
I'll change that one PR

cbosdo (approved):

PR #32, opened 2023-09-12
Create and configure logger

rjmateus:
Add support for logging

This is a multi-log, it will write the log to the console and also for a log file located at "/var/log/uyini-tools". This is useful to track what was called by the user.

Implemented for both uyuniadm and uyunictl tools.

This PR adds the logging configuration and migrate all the old log call and prints to a call in the new log, it doesn't check if we have enough log information.


cbosdo (commented):
Looks cool. Now I think we'll have to go through all the `log.Print*` calls and convert them

shared/utils/exec.go
```diff
@@ -43,7 +43,7 @@ func Exec(globalFlags *types.GlobalFlags, backend string, interactive bool, tty
 	}
 	commandArgs = append(commandArgs, "sh", "-c", strings.Join(args, " "))
 	if globalFlags.Verbose {
-		fmt.Printf("> Running: %s %s\n", command, strings.Join(commandArgs, " "))
+		log.Info().Msgf(" Running: %s %s", command, strings.Join(commandArgs, " "))
```
cbosdo, line 46:
```suggestion
		log.Info().Msgf("Running: %s %s", command, strings.Join(commandArgs, " "))
```

cbosdo (approved):
Tiny suggestion to remove a now unneeded white space in a log message.

PR #36, opened 2023-09-14
only print log caller in debug or lower level

rjmateus:
Improve logs for console log. 

at the console we only put the time: 
image

The file log will contain full date/time:
image

The caller method will only be printed if the log level is debug or lower, and will be printed to the console and log file:
image

image

### DIFF
--- a/shared/utils/logUtils.go
+++ b/shared/utils/logUtils.go
@@ -3,7 +3,6 @@ package utils
 import (
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


shared/utils/logUtils.go
```diff
@@ -38,6 +37,9 @@ func SetLogLevel(logLevel string) {
 	if err != nil {
 		level = zerolog.InfoLevel
 	}
+	if level <= zerolog.DebugLevel {
+		log.Logger = log.Logger.With().Caller().Logger()
```
cbosdo, line 41: Don't we want the caller always in the log file?

rjmateus: For that is all or nothing, the logging framework doesn't allow this or the caller will be the framework internal file.

PR #37, opened 2023-09-15
Add log level differenciation between console and file

cbosdo:
uyunictl exec needs only the output of the executed command by default but we may want to log some things to the file. By default the console logger is set to warn level and the file to info.


rjmateus (approved):
LGTM

PR #39, opened 2023-09-18
kubernetes: store SSL CA in database

cbosdo:
Kickstart feature requires the SSL CA to be loaded in the database. This is done by the setup script in the podman case and is already in the database after the migration.

We only need to add it to the database when installing on a kubernetes cluster.


mbussolotto (approved):
LGTM

PR #40, opened 2023-09-18
Add command for copying ISO images v2

aaannz:
Based on @nadvornik [PR#10](https://github.com/uyuni-project/uyuni-tools/pull/10)

- Instead of `distcp` use `distribution copy` command
- Aliases distribution - distro and copy - cp
- Automatically mount ISO before copying if .iso is specified as a source
- Rebase on current HEAD, including RunRawCmd refactor
- gofmt run


shared/utils/cp.go
```diff
@@ -42,3 +42,22 @@ func Copy(globalFlags *types.GlobalFlags, backend string, src string, dst string
...
+	switch command {
+	case "podman":
+		commandArgs = append(commandArgs, "test", "-e", dstpath)
+	case "kubectl":
+		commandArgs = append(commandArgs, "-c", "uyuni", "test", "-e", dstpath)
+	default:
+		log.Fatal().Msgf("Unknown container kind: %s\n", command)
+	}
+
+	if err := RunRawCmd(command, commandArgs, true); err != nil {
```
cbosdo, line 59: This will log the output to the console. I would use the new `RunCmdOutput()` and discard the output. If the file doesn't exist the function will get the proper return anyway

shared/utils/pod.go
```diff
@@ -3,31 +3,32 @@ package utils
...
+	"var-lib-cobbler":       "/var/lib/cobbler",
+	"var-pgsql":             "/var/lib/pgsql",
+	"var-cache":             "/var/cache",
+	"var-spacewalk":         "/var/spacewalk",
+	"var-log":               "/var/log",
+	"srv-salt":              "/srv/salt",
+	"srv-www-pub":           "/srv/www/htdocs/pub",
+	"srv-www-cobbler":       "/srv/www/cobbler",
+	"srv-www-osimages":      "/srv/www/os-images",
+	"srv-www-distributions": "/srv/www/distributions",
```
cbosdo, line 15: This volume will not be needed once https://github.com/SUSE/spacewalk/issues/22539 is done

uyunictl/cmd/cp/cp.go
```diff
@@ -32,8 +32,8 @@ func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
 		},
 	}
 
-	cpCmd.Flags().String("user", "", "User or UID to set on the destination file")
-	cpCmd.Flags().String("group", "", "Group or GID to set on the destination file")
+	cpCmd.Flags().String("user", "tomcat", "User or UID to set on the destination file")
+	cpCmd.Flags().String("group", "susemanager", "Group or GID to set on the destination file")
```
cbosdo, line 36: This would be the default for all `cp` commands. I wouldn't have them here

uyunictl/cmd/distro/distro.go
```diff
@@ -0,0 +1,96 @@
...
+		Args:    cobra.ExactArgs(2),
+		Aliases: []string{"cp"},
+		Run: func(cmd *cobra.Command, args []string) {
+			viper := utils.ReadConfig(globalFlags.ConfigPath, "ctlconfig", cmd)
+			if err := viper.Unmarshal(&flags); err != nil {
+				log.Fatal().Err(err).Msg("Failed to unmarshall configuration")
+			}
+			distCp(globalFlags, flags, cmd, args[1], args[0])
+		},
+	}
```
cbosdo, line 42: Ideally `cp`  command definition should go in its own file in `uyunictl/cmd/distro/cp/cp.go` to match the other commands' scheme. This will make it more readable when we'll add `distribution delete` command and friends.

uyunictl/cmd/distro/distro.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := utils.RunRawCmd("/usr/bin/sudo", umount_cmd, true); err != nil {
+		log.Fatal().Err(err).Msgf("Unable to unmount iso file, leaving %s intact", mountpoint)
+	}
+
+	os.Remove(mountpoint)
+}
+
+func distCp(globalFlags *types.GlobalFlags, flags *flagpole, cmd *cobra.Command, distroName string, source string) {
+	log.Info().Msgf("Copying distribution %s\n", distroName)
+	if _, err := os.Stat(source); errors.Is(err, os.ErrNotExist) {
```
cbosdo, line 63: There is a `FileExists()` function in `shared/utils/utils.go` that can be used instead of this.

uyunictl/cmd/distro/distro.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	srcdir := source
+	if strings.HasSuffix(source, ".iso") {
+		log.Debug().Msg("Source is an iso file")
+		tmpdir, err := os.MkdirTemp("", "uyuni-tools")
+		if err != nil {
+			log.Fatal().Err(err)
+		}
+		srcdir = tmpdir
```
cbosdo, line 79: We have so many places where we do this that we may think about creating a helper function for it

uyunictl/cmd/distro/distro.go
```diff
@@ -0,0 +1,96 @@
...
+			"-o", "ro,loop",
+			source,
+			srcdir,
+		}
+		if err := utils.RunRawCmd("/usr/bin/sudo", mount_cmd, true); err != nil {
+			log.Fatal().Err(err).Msg("Unable to mount iso file. Mount manually and try again")
+		}
+	}
+
+	utils.Copy(globalFlags, flags.Backend, srcdir, "server:"+dstpath, "tomcat", "susemanager")
```
cbosdo, line 93: Since you have the `tomcat:susemanager` owner here, the defaults are not needed in the `cp` command

cbosdo (commented):
Some code could be reused / rearranged there, but the rest looks good to me. As a bonus, or in a follow up PR you could add the distro to the DB.

uyunictl/cmd/cp/cp.go
cbosdo: I don't think we need to make it mandatory, unless we find out it leads to errors.

shared/utils/pod.go
cbosdo: There are only a few things on the way for this, but yes: that would make the volumes management way simpler.

cbosdo (approved):

PR #42, opened 2023-09-18
change log flag name

rjmateus:
change name for log level flag to follow the same pattern as the others


aaannz:
OTOH we have `logLevel` with ISSv2

cbosdo (commented):
I thought about this the other day but reverted to your original version because of the viper processing: the `-` would be mapped to `log.debug` property in the configuration. I'm not sure if we want this or just `logLevel`.

rjmateus:
agree with you, I will remove the PR

PR #49, opened 2023-09-22
Build vars

cbosdo:
Allow tweaking default values at build time. This is useful to set default according to where the package is built in OBS.


rjmateus (commented):
The only think that I see missing in how to build locally and affect the values

cbosdo:
> The only thing that I see missing in how to build locally and affect the values

Locally the uyuni stable defaults are used unless you pass the proper cflags... but do we want to make it so easy to tweak the local builds? OBS builds would be enough for me

rjmateus (approved):

rjmateus:
At this stage I don't know if is needed, since users can always pass the image flag when testing locally, but It should not produce any problem if we document how to do it

PR #52, opened 2023-09-25
hide debug flag, since is for developer only

rjmateus:
Hide debug flag, since it should not be presented for end users, it should be used for developers only.


rjmateus:
Closing for now, if we get questions from users about this I will re-create it and merge.

PR #53, opened 2023-09-25
uyuniadm subcommands

cbosdo:
Split `uyuniadm install` command into `uyuniadm install podman` and `uyuniadm install kubernetes` commands. This will make them clearly separated for conditional build in the future.


cbosdo:
Note: the function adding missing parameters seems to be broken

README.md
```diff
@@ -10,6 +10,17 @@
 `go build -o ./bin ./...` will produce the binaries in the root folder.
 Alternatively, if you have `podman` installed you can run the `build.sh` script to build binaries compatible with SLE 15 SP4 or openSUSE Leap 15.4.
 
+### Building in Open Build Service
```
rjmateus, line 13: this doesn't look to belong here

uyuni-tools.spec
```diff
@@ -57,7 +56,27 @@ tar -zxf %{SOURCE1}
 %build
 export GOFLAGS=-mod=vendor
 %goprep %{provider_prefix}
-mkdir -p bin && go build -o ./bin ./...
+mkdir -p bin
+ADM_PATH=%{provider_prefix}/uyuniadm/shared/utils
+
+tag=%{!?_default_tag:latest}
```
rjmateus, line 62: this is also part of other PR

uyuniadm/cmd/install/shared/flags.go
```diff
@@ -0,0 +1,100 @@
...
+	IssParent  string
+	MirrorPath string
+	Tftp       bool
+	Db         DbFlags
+	ReportDb   DbFlags
+	Cert       cmd_utils.SslCertFlags
+	Scc        SccFlags
+	Debug      bool
+	Image      cmd_utils.ImageFlags `mapstructure:",squash"`
+	Podman     cmd_utils.PodmanFlags
```
rjmateus, line 41: These podman and helm flags should be specific to the in use and not the share install flags

uyuniadm/cmd/install/shared/flags.go
```diff
@@ -0,0 +1,100 @@
...
+	cmd.Flags().String("cert-city", "Nuernberg", "SSL certificate city")
+	cmd.Flags().String("cert-org", "SUSE", "SSL certificate organization")
+	cmd.Flags().String("cert-ou", "SUSE", "SSL certificate organization unit")
+	cmd.Flags().String("cert-password", "", "Password for the CA certificate to generate")
+	cmd.Flags().String("cert-email", "ca-admin@example.com", "SSL certificate E-Mail")
+
+	cmd.Flags().String("scc-user", "", "SUSE Customer Center username")
+	cmd.Flags().String("scc-password", "", "SUSE Customer Center password")
+
+	cmd.Flags().Bool("debug", false, "Enable debugging features")
```
rjmateus, line 98: Sorry to get back to this, but we need to remove this flag or hide it with `cmd.Flags().MarkHidden("debug")`.
No customer or uyuni user will ever debug the code, and this will bring by providing a flag that a sub-set of people sometimes uses for development.

rjmateus (changes requested):

uyuniadm/cmd/install/shared/flags.go
cbosdo: I have already told you that I don't agree with your approach and this is not a reason to remove such a useful flag!

uyuni-tools.spec
cbosdo: I simply forgot to fetch `main`

cbosdo:
> Note: the function adding missing parameters seems to be broken

Fixed: just had to squash the mapstructure for the anonymous type.

rjmateus (approved):

PR #61, opened 2023-10-02
Split in two packages

cbosdo:
Separated uyuniadm and uyunictl in their own packages since those should be installed on different machines.

**Before merging this, a sumaform PR will be required to install the two packages instead of the the old `uyuni-tools`.**

### DIFF
--- a/uyuni-tools.spec
+++ b/uyuni-tools.spec
@@ -48,6 +48,21 @@ BuildRequires:       libgpgme-devel
 %description
 Tools for managing uyuni container.
 
+%package -n uyuniadm
+Summary:      Command line tool to install and update Uyuni
+
+%description -n uyuniadm
+uyuniadm is a convenient tool to install and update Uyuni components as containers running
+either on podman or a kubernetes cluster.
+
+%package -n uyunictl
+Summary:      Command line tool to perform day-to-day operations on Uyuni
+
+%description -n uyunictl
+uyunictl is a tool helping with dayly tasks on Uyuni components running as containers
+either on podman or a kubernetes cluster.
+
+
 %prep
 %autosetup
 tar -zxf %{SOURCE1}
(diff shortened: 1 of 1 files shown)


cbosdo:
@deneb-alpha would this help shipping the uyuni-tools in the proper channels?

rjmateus (approved):

deneb-alpha:
> @deneb-alpha would this help shipping the uyuni-tools in the proper channels?

having one source or two doesn't change a lot from a rel-eng POV. the work to be done is almost the same in terms of channels and product definition changes. 
But from end-user POV having the 2 subpackages I think it's better and will require to install only what is really needed for them.

uyuni-tools.spec
deneb-alpha, line 1: @cbosdo is there any reason for pinning the Go version in this way?

~~~
%if 0%{?rhel}
BuildRequires:  golang >= 1.19
%else
BuildRequires:  golang(API) = 1.19
%endif
~~~

cbosdo: I think this has been copy/pasted from another project. I can change that if needed indeed

cbosdo:
> > @deneb-alpha would this help shipping the uyuni-tools in the proper channels?
> 
> having one source or two doesn't change a lot from a rel-eng POV. the work to be done is almost the same in terms of channels and product definition changes.

Yes, but would that make it easier to get `uyunictl` in the client tools or another similar project building for many distros? `Uyuniadm` could then be shipped in a different place (to be figured out which, even for Uyuni)

uyuni-tools.spec
cbosdo: Fixed

aaannz:
Since uyuni-tools were not released yet, we don't need anything wrt. to provides-obsoletes combo, right?

deneb-alpha (approved):
LGTM

PR #63, opened 2023-10-05
Static build

cbosdo:
Statically compile for easier use on other distros.


rjmateus (approved):

cbosdo:
I merged the script part and left this PR opened until I have an answer from the OBS team

cbosdo:
For now we are building for all distros in OBS so this is not needed

PR #68, opened 2023-10-11
Define the goals and boundaries of each module

rjmateus:
The goal for this PR is continue the discussion about the goal/stakeholder and boundaries of each module provided by the uyuni-tools repository.
@cbosdo @juliogonzalez @admd @mmmmmmpc please feel free to contribute

### DIFF
--- a/modules.md
+++ b/modules.md
@@ -1,0 +1,60 @@
+The goal of this content is to set a high-level overview of each tool available.
+
+For tools that depend on the backend we should explicitly specify which one we want to use. Backend can also be defined in the configuration file to be used by the user.
+
+## Tools definition
+
+In case one wants to add a new sub-command it should decide to in which tool it should be placed.
+
+If the new sub-command needs access to the host OS of direct access to a running container then it should be added to MGRADM.
+
+Commands in MGRCTL should use the API only. 
+
+Any command to manage the proxy deployment must be placed in MGRPROXY.
+
+MGRDEV is focused on utility commands to be used during the development process.
+
+
+## MGRADM
+
+**Goals and definition:**
+
+Install, update, and maintain a containerized Uyuni Server. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
+
+Any new command that needs direct access to the host OS or any running container must be added to these tools.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** System where Uyuni Server should be deployed
+**Sub-commands Naming:** verb -> backend
+
+## MGRCTL
+
+**Goals and definition:**
+Helper tool for day-to-day operations and integration with other tools.
+Sub-commands in this tool should use the API calls (although case-by-case exceptions can be considered if there are valid reasons).
+
+**Target Stakeholder:** Uyuni operators
+**Where to install:** System where the Uyuni Server is deployed, or in the operator machine (supporting the same Operating Systems we already support for `spacecmd`).
+**Sub-commands Naming:** subcommand -> verb
+
+## MGRDEV
+
+**Goals and definition:**
+Utility commands to be used during development process. This tool can have commands that run remotely on the host OS or on running containers. These commands can use SSH and podman-socket.
+Examples of sub-commands are `cp` and `exec`.
+
+**Target Stakeholder:** Uyuni Developers
+**Where to install:** Any machine that needs remote access to running containers.
+**Sub-commands Naming:** verb -> backend
+
+
+## MGRPROXY
+
+**Goals and definition:**
+Install and manage a containerized Uyuni Proxy. This new command is a proposal to solve the problem of managing the Proxy using the same tool that manages the server, and how that can lead to confusion and errors.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** System where the Uyuni Proxy should be deployed
+**Sub-commands Naming:** verb -> backend
+
+This command is to be developed in a later stage since it would be better to redefine how we deploy containerized proxy and follow the same approach we have provided in the server.


modules.md
```diff
@@ -0,0 +1,42 @@
...
+
+## UYUNICTL
+
+**Goals and definition:**
+Helper tool for day to day operations. Sub-commands in this tool should preferably use the API.
+
+Some exceptions can be added to help administrators. One example is the `distro` sub-command, which is copying files inside the container and calling the API. Because it needs access to a container, if the tool is installed outside the HOST OS the podman-socket needs to be started and the command caller needs to have ssh access to the machine.
+
+TODO: 
+  - discuss if we should keep `cp` and `exec` sub-commands in here or if we should move them to a new tool `uyunidev` which would allow remote execution of it?
```
cbosdo, line 29: Note that in any case those could be grouped as Admin / debugging subcommands in the help

modules.md
```diff
@@ -0,0 +1,42 @@
+The goal for this content is to set a high-level overview of how to contribute and information about each tools available.
```
cbosdo, line 1: For all tool and command we probably will need to document the config file format from the possible options. `kubectl` has a hidden `options` command for global flags... do we want to be inspired by this for the config? Do we want a section in each command's help?

modules.md
```diff
@@ -0,0 +1,42 @@
...
+## Sub-command definition
+
+How should the sub-commands be organized at a high level. Currently we are using:
+- uyuniadm: verb -> backend
+- uyuniproxy?: verb -> backend
+- uyunictl: action -> verb
+
+# Existing Commands
+
+## UYUNIADM
```
juliogonzalez, line 12: Nitpick: is it too late to call this `uyuniserver`? I am telling because if we'll manage the Proxy with a separate tool (`uyuniproxy`) it could confuse some users... as I guess `adm` comes from administrate/administrator?

modules.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Goals and definition:**
+
+Install, update and mantain a uyuni server instalation. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
```
juliogonzalez, line 16:
```suggestion
Install, update and mantain a containerized Uyuni Server. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
```

modules.md
```diff
@@ -0,0 +1,42 @@
+The goal for this content is to set a high-level overview of how to contribute and information about each tools available.
+
+## Sub-command definition
+
+How should the sub-commands be organized at a high level. Currently we are using:
+- uyuniadm: verb -> backend
+- uyuniproxy?: verb -> backend
+- uyunictl: action -> verb
```
juliogonzalez, line 8: I am not sure I am understanding what the differences are here, but if this is clear for everyone else, no problem from my side :-)

modules.md
```diff
@@ -0,0 +1,42 @@
...
+# Existing Commands
+
+## UYUNIADM
+
+**Goals and definition:**
+
+Install, update and mantain a uyuni server instalation. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** OS where server should be deployed
```
juliogonzalez, line 19:
```suggestion
**Where to install:** System where Uyuni Server should be deployed
```

modules.md
```diff
@@ -0,0 +1,42 @@
...
+Helper tool for day to day operations. Sub-commands in this tool should preferably use the API.
+
+Some exceptions can be added to help administrators. One example is the `distro` sub-command, which is copying files inside the container and calling the API. Because it needs access to a container, if the tool is installed outside the HOST OS the podman-socket needs to be started and the command caller needs to have ssh access to the machine.
+
+TODO: 
+  - discuss if we should keep `cp` and `exec` sub-commands in here or if we should move them to a new tool `uyunidev` which would allow remote execution of it?
+  - should we also move `distro` command to uyuniadm or think of a way to run it through the api?
+
+**Target Stakeholder:** Uyuni operator and administrator
+**Where to install:** On podman Host OS or in the operator machine, supporting the same OS's we already support for spacecmd.
```
juliogonzalez, line 33:
```suggestion
**Where to install:** System where the Uyuni Server is deployed, or in the operator machine (supporting the same Operating Systems we already support for `spacecmd`).
```

modules.md
```diff
@@ -0,0 +1,42 @@
...
+
+
+## UYUNIPROXY
+
+**Goals and definition:**
+Install and manage containerized proxy. This new command is a proposal to solve the problem of managing the proxy using the same tool that manages the server, and how that can lead to confusion and errors.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** OS where proxy should be deployed
```
juliogonzalez, line 42:
```suggestion
**Where to install:** System where the Uyuni Proxy should be deployed
```

modules.md
```diff
@@ -0,0 +1,42 @@
...
+  - should we also move `distro` command to uyuniadm or think of a way to run it through the api?
+
+**Target Stakeholder:** Uyuni operator and administrator
+**Where to install:** On podman Host OS or in the operator machine, supporting the same OS's we already support for spacecmd.
+
+
+## UYUNIPROXY
+
+**Goals and definition:**
+Install and manage containerized proxy. This new command is a proposal to solve the problem of managing the proxy using the same tool that manages the server, and how that can lead to confusion and errors.
```
juliogonzalez, line 39:
```suggestion
Install and manage a containerized Uyuni Proxy. This new command is a proposal to solve the problem of managing the Proxy using the same tool that manages the server, and how that can lead to confusion and errors.
```

modules.md
juliogonzalez: Real question here is, do we want `uyunictl` users to be able to call `cp` and `exec`? If answer is yes... do we want it even if they are not using API calls?

If the answer to any of the questions above is no -> `uyunidev`
If the answer to both questions above is yes -> `uyunictl` and maybe a group of subcommands as suggested by @cbosdo

modules.md
```diff
@@ -0,0 +1,42 @@
...
+
+Install, update and mantain a uyuni server instalation. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** OS where server should be deployed
+
+## UYUNICTL
+
+**Goals and definition:**
+Helper tool for day to day operations. Sub-commands in this tool should preferably use the API.
```
juliogonzalez, line 24: I'd define this as:
```suggestion
Helper tool for day to day operations. Sub-commands in this tool must use the API (although case-by-case exceptions can be considered if there are valid reasons).
```
Way less open to interpretation. If someone wants to apply for an exception, it should be very clear why.

modules.md
rjmateus: From my point of view it would be better to have the `cp` and `exec` commands in the uyuniadm, since it's more restricted in terms of users and usage. Then if we find the need to open the scope we can easily do it. 
The other way around, add it to uyuctl and then find it too open, it would be more difficult to reduce the scope and make it more restrict.
And I agree with you, if we move in the direction of putting those in the uyuniadm, then we need to have a uyunidev with remove useful commands for developer.

modules.md
```diff
@@ -0,0 +1,42 @@
...
+
+Some exceptions can be added to help administrators. One example is the `distro` sub-command, which is copying files inside the container and calling the API. Because it needs access to a container, if the tool is installed outside the HOST OS the podman-socket needs to be started and the command caller needs to have ssh access to the machine.
```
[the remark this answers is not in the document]
rjmateus: I would say that will depend if we want to have remote commands on the uyunictl

modules.md
rjmateus: we would need to experiment on this one.
Other think we can think about is a default location for the config file, that if it exists we read it, and avoid the need to pass the  `--config` parameter

modules.md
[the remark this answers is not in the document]
juliogonzalez: Exactly, but FMPOV that's something we need to define now. See the other comment I left about what we should have in `uyunictl` :-)

modules.md
cbosdo: We already have a default location and name... but the name doesn't please me that much for now

modules.md
cbosdo: Honestly I don't find `uyuniserver` and `uyuniproxy` names that nice... maybe I would be more inclined in getting all into `uyuniadm`

modules.md
cbosdo: No I don't get the `action -> verb` either 😉

modules.md
juliogonzalez: That's another option. But if we finally go with separate commands, then I think it should be clear that the one for the server is only for the server. Just my 2 cents :-)

rjmateus: we still have time. Uyunictl to control the server only? Or should we also add some ability to control proxy? (it doesn't have a API, so it should be hard).
But is something to think about

modules.md
rjmateus: Me neither :face_with_spiral_eyes: 

I think I was thinks something like "uyunictl channel create/delete/get .."

modules.md
aaannz: IIUC we'll need `cp` to gather support config from containers, won't we? So "hiding" it in `uyunidev` seems not troubleshooting friendly. I'm all for having it in `uyuniadm`.

modules.md
aaannz: What is the default location? Just `$HOME/.config` as is in [config.go](https://github.com/uyuni-project/uyuni-tools/blob/main/shared/utils/config.go#L17)? In my [PR](https://github.com/uyuni-project/uyuni-tools/pull/71/files#diff-5cb9cbb70541bd2cb439f880704af2a4d133f6ca9f2661844f4ccb139bda5307R12) I added `/usr/share/uyuniadm/` for rpm shipped configs and `/etc/uyuni/` for admin configs. Let me know where to change it.

modules.md
juliogonzalez: Valid point, but in that case: shouldn't we remove it from `uyunictl` and have it at `uyunidev` and `uyuniadm`. Or maybe I am missing something, but I thought we wanted to avoid having developers using `uyuniadm` on their own laptops, and what's why `uyunidev` would be something nice to have.

modules.md
aaannz: My opinion, or rather preference, is to have only one command - `uyuniadm` and have a distinction after command verb - `uyuniadm install proxy`, `uyuniadm install server`.
We can write down in a cache file what we installed, or can evaluate installed containers so that if distinctive noun is missing, we by default do actions on installed variant. If both are installed on one host, we require noun. Not sure how easily it can be to do in cobra.

juliogonzalez: > we still have time. Uyunictl to control the server only? Or should we also add some ability to control proxy? (it doesn't have a API, so it should be hard). But is something to think about

Fair question.

But while the proxy doesn't have an API, don't we have API calls for the server that are used for interactions with the proxy? For example: https://www.uyuni-project.org/uyuni-docs-api/uyuni/api/proxy.html Would all those API calls be valid only for `uyuniproxy`?

If in the end we want users from outside the server/proxy to interact with the server/proxy, I'd have a single tool `uyunictl` for that. But then I wonder if we don't want the same for `uyuniadmin`, so it can handle both Server and Proxy :-)

modules.md
rjmateus: in uyuniadm we would be able to use cp and exec without the need to start the podman socket, since we are running it on the host OS machine.
For developers we need to be able to run it remotely, so I would say it's needed a uyunidev application to allow it. 
With this in mind I think we can have it on uyuniadm and uyunidev, if we remove them from uyunictl.

modules.md
rjmateus: If we want to read properties from those locations we need to add them to vipper configuration. I would say we don't need to add rpm shipped configuration since the defaults will be on the code and we are talking about a command line tool, and not a web application.

modules.md
aaannz: `uyunictl` does not have `backend`.
But I agree that not all commands have backend. Like my `uyuniadm distro copy ...` does not have `backend` option and tries to parse if from config file or mere presence of the tools.
So IMO `backend` is only valid for `install` command.

modules.md
aaannz: IMO proxy should be not interacted with at all apart of debugging and start/stop.
`uyunictl` is for server only.

rjmateus: @juliogonzalez the api calls you share influence the proxy but are made to the server API. Proxy doesn't have any direct API available.

I like @aaannz proposal. however have that kind of dynamic verb/noun it's not simple. We would need to run a bunch of code to detect environment before compose cobra command (but I think it's possible to implemen).

Currently, we are also making the distinction between podman and kubernetes management by putting it as a sub-command. Example "uyuniadm install podman"
Ondrej you are  proposing something similar to "uyuniadm install podman server".

modules.md
cbosdo: we somehow need to be able to force the selection of the backend since we cannot fully rely on the `kubectl` / `podman` detection when both are available

modules.md
cbosdo: For the support config we would probably have a dedicated command in `uyuniadm`: we want to gather the supportconfig from within the container but also data from the host.

modules.md
agraul: For rpm-shipped configs, if we want them, isn't [`/usr/etc`](https://en.opensuse.org/openSUSE:Packaging_UsrEtc) current location in (open)SUSE?

modules.md
agraul: Maybe we could use `subcommand` instead of `action`? Action is a bit unfortunate IMO because we use it with a different meaning in Uyuni itself.

modules.md
agraul: > We would need to run a bunch of code to detect environment before compose cobra command (but I think it's possible to implement).

Can you expand on this a bit? What part do we need to compute before composing the cobra command?

> Currently, we are also making the distinction between podman and kubernetes management by putting it as a sub-command. Example "uyuniadm install podman"

Should this maybe be an --adjective/--flag?

modules.md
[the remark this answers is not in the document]
agraul: [nitpick] `s/inside/into/`

modules.md
cbosdo: I don't think we want to provide any config by default with the tools

modules.md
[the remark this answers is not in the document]
cbosdo: > Is there any downside to enabling the `podman.socket`? Systemd sockets have a low overhead afaik.

The only downside, if that is any, is that users can remote connect to podman over SSH super easily. To me there is no downside at all.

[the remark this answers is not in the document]
rjmateus: I agree with Cedric, I don't see a problem with enabling it by default. 
I put this statement in here just to make it clear that if we want to add remote commands to uyunictl, then the podman.socket enabled is required.

cbosdo:
As a test for these rules: where should we add a command to generate the proxy configuration and how should it be called? This requires access to the server container for `podman`  and the cluster for kubernetes since we have to generate SSL certificate pair.

modules.md
juliogonzalez: Aha, if we do it according to that definition, `should` is OK. But not everyone knows about the definition on the RFC or agrees with it so what about:
```suggestion

Helper tool for day to day operations. Sub-commands in this tool should use the API (although case-by-case exceptions can be considered if there are valid reasons).
```

modules.md
admd: `uyuniadm install podman server ` -> layman question - this is getting complex. Would there be some suggestions offered to the user on each level?

admd: If things are getting complex for doing all from single tool which although wound be nice,  having a separate one one for proxy wouldn't be a disaster either.

admd: `Should this maybe be an --adjective/--flag?` +1

modules.md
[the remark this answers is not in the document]
admd: If we don't enable it by default, what would user needs to do to enable it? Should we not disable it by default and if user need it, they can enable it following the documentation that we provide. Here I am assuming it's straightforward to enable it.

modules.md
admd: Personally, I am not in favor of uyunidev, we should be using tools that our end user will be using. It will also help us identify pain points of our users.

modules.md
cbosdo: > `uyuniadm install podman server ` -> layman question - this is getting complex. Would there be some suggestions offered to the user on each level?

The nice thing is that `cobra` provides shell completion for free. I haven't looked at how those work, but should make it easier with the suggestions

modules.md
rjmateus: I think we don't need a default config file, since it would be all customer-specific. 
Regarding the options suggested by Cedric, I think it's an implementation detail that can be good to have.

modules.md
rjmateus: @agraul the way we organized cobra sub-commands are computed based on levels. Conditional present one level would force to re-implement the structure of our sub-commands.
You can see it in here: https://github.com/uyuni-project/uyuni-tools/blob/main/uyuniadm/cmd/install/install.go

modules.md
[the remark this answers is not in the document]
cbosdo: The only thing to be done is `systemctl enable podman.socket` on the host.

rjmateus:
I have updated the content with the feedback. Please let me know what you think about it.

modules.md
mmmmmmpc: I advocate for no initial config file and to use a `$HOME/.mgr/config` folder (following the use of `mgr` in naming we agreed on)

modules.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Tools definition
+
+In case one wants to add a new sub-command it should decide to in which tool it should be placed.
+
+If the new sub-command needs access to the host OS of direct access to a running container then it should be added to UYUNIADM.
+
+Commands in UYUNICLT should use the API only. 
+
+Any command to manage the proxy deployment must be placed in UYUNIPROXY.
```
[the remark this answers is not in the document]
rjmateus: I was giving another thought to this one, and I think it should be a separated command. We have a chat about it, but high level the reasons are:
- We do not expect to install a server and proxy on the same machine, so we should not need to have both at the same time.
- With different tools, we can get a reduced scope in each one, and avoid mistakes from the user
- Avoid another level in the command call
- All code can be shared internally since both are placed in the same repository

modules.md
```diff
@@ -0,0 +1,60 @@
...
+
+UYUNIDEV is focused on utility commands to be used during the development process.
```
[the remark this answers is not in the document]
rjmateus: The goal for this tool is to be able to call exec and cp remotely to better integrate with the existing ant scripts and testsuite. We should not expect much more then that. 
But your concern are valid, and if we move forward with this we need to keep an eye on it.

modules.md
```diff
@@ -0,0 +1,60 @@
...
+
+**Goals and definition:**
+
+Install, update, and maintain a containerized Uyuni Server. Commands placed here will have/need access to the container runtime environment and also to the HOST OS.
+
+Any new command that needs direct access to the host OS or any running container must be added to these tools.
+
+**Target Stakeholder:** Uyuni administrator
+**Where to install:** System where Uyuni Server should be deployed
+**Sub-commands Naming:** verb -> backend
```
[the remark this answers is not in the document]
rjmateus: Those are great questions. Should we always have the backend definition on the sub-command?
Or should we try to have an intelligent guess? For example, we try to detect podman and k*s configuration. If we find just one, we continue with that. If we found both we ask which one to use (and add a flag to specify it).
Since the distro is already implemented with the file copy I think we should keep it. In future, if we implement it to use the API to deal with that and move it to uyunictl? If we have a delete command, since it would only use the API, should we place it in uyunictl? Right now we are trying to mimic the existing user access restrictions to try to avoid be to disruptive.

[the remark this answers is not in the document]
cbosdo: > Since the distro is already implemented with the file copy I think we should keep it. In future, if we implement it to use the API to deal with that and move it to uyunictl? If we have a delete command, since it would only use the API, should we place it in uyunictl? Right now we are trying to mimic the existing user access restrictions to try to avoid be to disruptive.

Better avoid moving such commands between the two tools, even in the future: that would only create confusion for users

modules.md
```diff
@@ -0,0 +1,60 @@
...
+
+For tools that depend on the backend we should explicitly specify which one we want to use. Backend can also be defined in the configuration file to be used by the user.
+
+## Tools definition
+
+In case one wants to add a new sub-command it should decide to in which tool it should be placed.
+
+If the new sub-command needs access to the host OS of direct access to a running container then it should be added to MGRADM.
+
+Commands in MGRCLT should use the API only.
```
aaannz, line 11:
```suggestion
Commands in MGRCTL should use the API only. 
```

aaannz (approved):

mmmmmmpc:
Looks good to me.
I think we are good to merge.

rjmateus:
I will merge this PR, since no more comments were added. Remember that this is not written in stone, and can be changed in any time in the future.

PR #70, opened 2023-10-11
Make fqdn optional for podman install

rjmateus:
If no FQDN is passed to the command it will look at the host full name by calling "hostmane -f".


uyuniadm/cmd/install/podman/utils.go
```diff
@@ -44,6 +45,9 @@ func pullImage(flags *podmanInstallFlags) {
 }
 
 func installForPodman(globalFlags *types.GlobalFlags, flags *podmanInstallFlags, cmd *cobra.Command, args []string) {
+	fqdn := getFqdn(args)
+	log.Debug().Msgf("seting up server with the fqdn '%s'", fqdn)
```
cbosdo, line 49:
```suggestion
	log.Info().Msgf("setting up server with the FQDN '%s'", fqdn)
```

cbosdo (commented):
Typo and log level change suggestion

PR #71, opened 2023-10-11
API module and distro auto-registration

aaannz:
Fixes https://github.com/SUSE/spacewalk/issues/22547

This PR implements:
1) basic API module to facilitate JSON HTTP access to the uyuni server

Uses `net/http` module, exports `Get` and `Post` helpers to send GET and POST requests and parse response to map[string]interface{}. Not sure if we want to export `sendRequest` as well so other parts can work directly over Response type.
Basic usage is for example
```go
client := api.Init(flags)
client.Login(flags.User, flags.Password)
client.Post(query_path, query_data)
```
I am open to adding `Login` directly to the `Init` and not exporting it. Initially I didn't pass complete flags to the `Init`, just fqdn, that's why this design.

Also adds helper for adding cmdline flags --server, --username, --password, --cacert, --insecure. Helper has an option to make API options optional or mandatory.

2) api command option to the uyunictl

Example usage `uyunictl api --server server --username user --password pass get image/profile/getDetails label=myimage`

Is is also possible to pass custom CA certificate using `--cacert $path_to_cacert`` option or skip verification using `--insecure` flag.

Maybe we can later add support for caching credentials or wallets to ease always typing these options.

3) Distribution auto-registration

`distro copy` command moved from `uyunictl` to `uyuniadm` as this requires to be run on container host.

If API details were supplied to the distro copy, after distribution is copied to the container, post query is send registering this distribution to the uyuni db.

Before it can be registered, uyuniadm needs to determine what distribution it is, installer type and channel label. ~~For this I added extra config file, by default installed into `/usr/share/uyuniadm/productmap.yaml`.~~
Data about this are hardcoded with basic distro-version lookup, but is also optional to add it to user's config file under `~/.config`. This maps distribution names, as they appear in distro `.treeinfo` file, to installer type and default channel label.

~~I opted to use config file instead of having this in the code so it can be easily extended without rebuilding. `productmap.yaml` can also be copied and modified by admin to the `/etc/uyuni/productmap.yaml`, as is the standard.~~

In case of SLE, it by default uses SLES channel, however that may not be right all the time, hence I added 3d optional parameter to `distro copy` command specifying what parent channel to use.


aaannz:
> Do you need to have the distos in a yaml file to allow overwrite? Could/should we move it to a const map go? The good part is that it would be compiled inside the code and no extra external file would be needed

I don't require it, I just found it easier to amend when there is a new distro instead having data in the code. That's what I initially had it in the code, but once I started to add more distributions, it was easier for me to have that in the separate file.

rjmateus:
> > Do you need to have the distos in a yaml file to allow overwrite? Could/should we move it to a const map go? The good part is that it would be compiled inside the code and no extra external file would be needed
> 
> I don't require it, I just found it easier to amend when there is a new distro instead having data in the code. That's what I initially had it in the code, but once I started to add more distributions, it was easier for me to have that in the separate file.

Ok I got your point. I think if the file is not expected to be changed or re-configure after installation (like the rhn.conf file) then it would be better to have it in the code.  This way be compiled and shipped inside the binary, and avoid someone changing it wrongly.
I it could be edited by customers, then it needed to be a sapareted file for sure.

uyuniadm/cmd/distro/distro.go
```diff
@@ -3,16 +3,19 @@ package distro
...
 
 type flagpole struct {
-	Backend string
+	Backend      string
+	ChannelLabel string
 }
 
 func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
 	flags := &flagpole{}
+	apiFlags := &api.ConnectionDetails{}
```
rjmateus, line 18: I'm not sure if you need to define the API flag in here or if you could set the flagpole to inherit also the api flags, and avoid this extra variable

rjmateus (commented):
just a few questions. Thank you for the great contribution

cbosdo:
> > > Do you need to have the distos in a yaml file to allow overwrite? Could/should we move it to a const map go? The good part is that it would be compiled inside the code and no extra external file would be needed
> > 
> > 
> > I don't require it, I just found it easier to amend when there is a new distro instead having data in the code. That's what I initially had it in the code, but once I started to add more distributions, it was easier for me to have that in the separate file.
> 
> Ok I got your point. I think if the file is not expected to be changed or re-configure after installation (like the rhn.conf file) then it would be better to have it in the code. This way be compiled and shipped inside the binary, and avoid someone changing it wrongly. I it could be edited by customers, then it needed to be a sapareted file for sure.

We could even imagine users being able to use a file to be merged with the hard-coded data. This way by default nothing is needed and fits in the binary, but advanced cases can still be covered.

aaannz:
I addressed raised issues:

* removed productmap.yaml, now some data are hardcoded, it is also possible to add product map settings to the admconfig user config file
* APIflags integrated with overall command flags and viper is used to marshal it
* added some comments for the shared API module and unexported `login` which is now automatically called during `Init`

shared/api/api.go
```diff
@@ -0,0 +1,250 @@
...
+	cmd.PersistentFlags().String("password", "", "Password for the API user")
+	cmd.PersistentFlags().String("cacert", "", "Path to a cert file of the CA")
+	cmd.PersistentFlags().Bool("insecure", false, "If set, server certificate will not be checked for validity")
+
+	if optional {
+		cmd.MarkFlagRequired("server")
+		cmd.MarkFlagRequired("username")
+		cmd.MarkFlagRequired("password")
+	}
+}
```
cbosdo, line 69: All the flags for API should be prefixed with `api-` to map them like this in the config:
```
api:
  server: ...
  user: ....
  ...
```

Is there a reason to use `PersistentFlags()` over `Flags()`?

uyuniadm/cmd/distro/distro.go
```diff
@@ -26,21 +30,31 @@ func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
...
+		Long: `Takes a path to iso file or directory with mounted iso and copies it into the container.
+
+Distribution name specifies the destination directory under /srv/www/distributions.
+
+Optional channel label specify which parent channel to associate with the distribution. Only when API details are provided and auto registration is done.`,
 		Args:    cobra.ExactArgs(2),
 		Aliases: []string{"cp"},
 		Run: func(cmd *cobra.Command, args []string) {
-			viper := utils.ReadConfig(globalFlags.ConfigPath, "ctlconfig", cmd)
+			viper := utils.ReadConfig(globalFlags.ConfigPath, "admconfig", cmd)
```
cbosdo, line 43: We probably want to have only one config file for all tools, but that would be in another PR

uyunictl/cmd/api/api.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
+	apiFlags := &api.ConnectionDetails{}
+
+	apiCmd := &cobra.Command{
+		Use:   "api",
+		Short: "JSON over HTTP API helper tool",
+	}
+
+	apiGet := &cobra.Command{
```
cbosdo, line 23: We probably should split the get and post into their own folder like other subcommands, but that is cosmetic and can be changed later

uyuniadm/cmd/distro/distro.go
```diff
@@ -7,12 +7,16 @@ package distro
...
+	"github.com/uyuni-project/uyuni-tools/shared/api"
 	"github.com/uyuni-project/uyuni-tools/shared/types"
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
 type flagpole struct {
-	Backend string
+	Backend      string
+	ChannelLabel string
+	ProductMap   map[string]map[string]types.Distribution
```
cbosdo, line 18: Where is the `ProductMap` defined? I don't see it in the command flags for the user to override it.

shared/api/api.go
aaannz: If I understood correctly, persistent flag allows me to do `mgradm distro copy --server --user ...` but also `mgradm distro --server --user`.
Same when used with `mgrctl api post`.

I only need to add flags to `distro` and `api` commands and are inherited to the child commands.

uyuniadm/cmd/distro/distro.go
aaannz: This is that viper can read it from `admconfig` file. Maybe the naming is not the best here.

shared/api/api.go
cbosdo: The problem is that those are not available as `Flags` in the cmd and if you pass them using CLI arguments that won't fly with the current code since it only gets the Flags into the configuration structure. Persistent flags are passed as GlobalFlags instead.

PR #72, opened 2023-10-12
Misc refactoring from airport

cbosdo:
Refactored the wait we get the pod name and split the `uyuniadm uninstall` command in two.


cbosdo:
Installation and migration have been successfully tested for podman and kubernetes. Uninstall has some annoying behavior to fix

cbosdo:
> @cbosdo what is the problem with un-install? At least with podman I have tested it and I'm using it and didn't found any issue so far

The problem happens when the install fails and both `kubectl` and `podman` are installed on the host: either uninstall picks `kubectl` as the connection works or it picks nothing if we check for a running container too. I'll probably add a backend parameter as on uyunictl to force the selection on one or the other for that particular case.

PR #76, opened 2023-10-19
Conditional build of uyuniadm

cbosdo:
Only build uyuniadm on recent distros. The others can have uyunictl for the client part though


deneb-alpha (commented):
the spec LGTM, but, what about the .changes?

cbosdo:
> the spec LGTM, but, what about the .changes?

Do we need a changelog entry for something that hasn't been released yet?

PR #77, opened 2023-10-19
Add script to automate pushing the package to OBS

cbosdo:


agraul (commented):
Question: Since the script takes an env var to set the API server, have you tested this with IBS? Will it work with ssh-key based authentication?

cbosdo:
> Question: Since the script takes an env var to set the API server, have you tested this with IBS? Will it work with ssh-key based authentication?

Hum... that is a good question. No I haven't tested that yet.

PR #91, opened 2023-10-27
Enable systemd colorized output

srbarrios:
I noticed that the commands inside our server container don't output with colors by default, so we need to enforce them manually. 

<details>
<summary>
Example:
</summary>

NO COLORS:
image
COLORS:
image
</details>

### DIFF
--- a/uyuniadm/shared/templates/serviceTemplate.go
+++ b/uyuniadm/shared/templates/serviceTemplate.go
@@ -20,6 +20,7 @@ RequiresMountsFor=%t/containers
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Environment=UYUNI_IMAGE={{ .Image }}
 Environment=TZ={{ .Timezone }}
+Environment=SYSTEMD_COLORS=1
 Restart=on-failure
 ExecStartPre=/bin/rm -f %t/uyuni-server.pid %t/%n.ctr-id
 ExecStartPre=/usr/bin/podman rm --ignore --force -t 10 {{ .NamePrefix }}-server


mbussolotto (approved):
LGTM

uyuniadm/shared/templates/serviceTemplate.go
```diff
@@ -20,6 +20,7 @@ RequiresMountsFor=%t/containers
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Environment=UYUNI_IMAGE={{ .Image }}
 Environment=TZ={{ .Timezone }}
+Environment=SYSTEMD_COLORS=1
```
cbosdo, line 23: I would probably not do this in the systemd unit file, but rather in the image itself.

cbosdo (changes requested):
After some tests, the systemd commands are getting colors when passing the `-ti` parameters to `uyunictl exec`. So I'm not sure this PR is needed at all as this mimics the systemd behavior: only get colors in interactive shells, not when called from scripts.

In any case to enforce colors without those parameters it would be better to do it with:

```diff
--- a/uyunictl/cmd/exec/exec.go
+++ b/uyunictl/cmd/exec/exec.go
@@ -69,7 +69,7 @@ func run(flags *flagpole, cmd *cobra.Command, args []string) {
                commandArgs = append(commandArgs, "-c", "uyuni", "--")
        }
 
-       newEnv := []string{}
+       newEnv := []string{"SYSTEMD_COLORS=1"}
        for _, envValue := range flags.Envs {
                if !strings.Contains(envValue, "=") {
                        if value, set := os.LookupEnv(envValue); set {
```

srbarrios:
In closing, the good practice is to use -ti

Issue #74, opened 2023-10-18
Shell auto-completion


cbosdo:
Cobra provides off-the-shelf auto-installation feature. This needs to be leveraged and packages as RPMs.

PR #94, opened 2023-11-06
Shell auto completion

NamelessOne91:


NamelessOne91:
Working on issue  #74: Shell auto-completion

I've added a shared cobra command  - completion -  to be used as sub-command both for uyuniadm and uyunictl root commands.
It leverages cobra's built in generation of shell completion scripts.

I had to do a minor refactor of the Logger's init method in order to avoid redirecting also the info welcome messages to the generated shell completion scripts.

build.sh
```diff
@@ -2,3 +2,11 @@
 
 mkdir -p ./bin
 go build -tags netgo -o ./bin ./...
+
+COMPLETION_FILE="/tmp/completion"
+
+# generate and source shell completion scripts for uyuniadm and uyunictl
+./bin/uyuniadm completion bash > "${COMPLETION_FILE}"
+./bin/uyunictl completion bash >> "${COMPLETION_FILE}"
```
mbussolotto, line 10: The two command should be run inside the spec file, in this way the completion would work out of the box installing uyuni-tools from RPM

mbussolotto (changes requested):

build.sh
cbosdo: Note that the completion files should be shipped in separate packages as done in this package for example: https://build.opensuse.org/package/view_file/devel:tools:scm/tig/tig.spec?expand=1

shared/completion/completion.go
```diff
@@ -0,0 +1,32 @@
...
+			switch args[0] {
+			case "bash":
+				cmd.Root().GenBashCompletion(os.Stdout)
+			case "zsh":
+				cmd.Root().GenZshCompletion(os.Stdout)
+			case "fish":
+				cmd.Root().GenFishCompletion(os.Stdout, true)
+			case "powershell":
+				cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
+			}
```
cbosdo, line 28: Do we need powershell completion since we are not building for windows? I would remove it.

uyuniadm/cmd/cmd.go
```diff
@@ -22,8 +24,15 @@ func NewUyuniadmCommand() *cobra.Command {
 	}
 
 	rootCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
+		utils.LogInit(true)
 		utils.SetLogLevel(globalFlags.LogLevel)
-		log.Info().Msgf("Executing command: %s", cmd.Name())
+
+		// do not log if running the completion cmd as the output is redirect
```
cbosdo, line 30:
```suggestion
		// do not log if running the completion cmd as the output is redirected
```

uyuniadm/cmd/cmd.go
```diff
@@ -22,8 +24,15 @@ func NewUyuniadmCommand() *cobra.Command {
...
+		// to create a file to source
```
cbosdo, line 31: Can't this line fit with the previous one? We cut at 120 characters, may be that fits

uyunictl/cmd/cmd.go
```diff
@@ -24,14 +25,21 @@ func NewUyunictlCommand() *cobra.Command {
...
 	rootCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
-		utils.LogInit("uyunictl", cmd.Name() != "exec")
+		utils.LogInit(cmd.Name() != "exec")
 		utils.SetLogLevel(globalFlags.LogLevel)
-		log.Info().Msgf("Executing command: %s", cmd.Name())
+
+		// do not log if running the completion cmd as the output is redirect
+		// to create a file to source
+		if cmd.Name() != "completion" {
+			log.Info().Msg("Welcome to uyuniadm")
```
cbosdo, line 34: Copy-paste issue:

```suggestion
			log.Info().Msg("Welcome to uyunictl")
```

build.sh
```diff
@@ -2,3 +2,11 @@
 
 mkdir -p ./bin
 go build -tags netgo -o ./bin ./...
+
+COMPLETION_FILE="/tmp/completion"
```
cbosdo, line 6: Why this path in the `build.sh` script? Can't those be in `bin` folder too?

README.md
```diff
@@ -50,6 +50,22 @@ https://github.com/uyuni-project/uyuni/tree/master/containers/doc/server-kuberne
...
+
+You'll then need to source the resulting script(s).
+
+As an example, to enable bash completion for uyuniadm:
+
+`./bin/uyuniadm completion bash > /tmp/completion`
+
+`. /tmp/completion`
+
+The supported shells are: bash, zsh, fish and powershell.
```
cbosdo, line 67: I would remove `powershell` from the list. See the other comment on that.

cbosdo (changes requested):
May be you should squash all the commits into one: those are not huge changes and are all related.

rjmateus:
Thank you for your contribution, this is really cool stuff.
However, I'm not sure if the completion subcommand should be part of the existing commands. What do you guys think?
We could have a module that could be called to generate the completion for all flavors and it could be called at build time. to provide the data for the completion package.
@cbosdo @mbussolotto what do you think?

mbussolotto:
@rjmateus yes, I think we summarized everything here https://github.com/uyuni-project/uyuni-tools/pull/94#discussion_r1386313269 (not really readable since it was a comment on the code)

cbosdo:
> Thank you for your contribution, this is really cool stuff. However, I'm not sure if the completion subcommand should be part of the existing commands. What do you guys think? We could have a module that could be called to generate the completion for all flavors and it could be called at build time. to provide the data for the completion package. @cbosdo @mbussolotto what do you think?

If that is possible, then yes. Otherwise we could just hide those subcommands.

cbosdo (approved):
LGTM now

uyunictl/cmd/cmd.go
cbosdo: > is this supposed to really be uyuniadm? can we use a variable here or it doesn't work?

Right! we should probably use `os.Args[0]` to make it future-proof

build.sh
```diff
@@ -12,3 +12,12 @@ offset=$(git rev-list --count ${tag})
...
 
 go build -tags netgo -ldflags "-X ${VERSION_NAME}=${tag}-${offset}" -o ./bin ./...
+go build -tags netgo -o ./bin ./...
+
+for shell in "bash" "zsh" "fish"; do
+    COMPLETION_FILE="./bin/completion.${shell}"
+
+    # generate and source shell completion scripts for uyuniadm and uyunictl
+    ./bin/uyuniadm completion ${shell} > "${COMPLETION_FILE}"
+    ./bin/uyunictl completion ${shell} >> "${COMPLETION_FILE}"
```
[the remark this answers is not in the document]
cbosdo: This is only for dev's use, but yes that's what it does.

uyuni-tools.spec
```diff
@@ -61,21 +75,108 @@ BuildRequires:  golang >= 1.19
...
+%description -n %{name_ctl}
+%{name_ctl} is a tool helping with dayly tasks on %{productname} components running as containers
 either on Podman or a Kubernetes cluster.
 
+%package -n %{name_adm}-bash-completion
+Summary:        Bash Completion for %{name_adm}
+Group:          System/Shells
+Requires:       %{name_adm} = %{version}
+%if 0%{?suse_version} >= 150000
+Supplements:    (%{name_adm} and bash-completion)
```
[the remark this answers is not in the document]
cbosdo: The build failed with it for SLE 12, Debian and Ubuntu: this is why I added the condition on SUSE versions above and including 15

cbosdo:
Note that this PR is built successfully for all distros but AlmaLinux8 in https://build.opensuse.org/package/show/home:cbosdonnat:branches:systemsmanagement:Uyuni:Master:ContainerUtils/uyuni-tools

mbussolotto (approved):
LGTM!

deneb-alpha (approved):
LGTM, thanks!

Issue #89, opened 2023-10-26
Installation fails after an uninstallation


mbussolotto:
Once I uninstall, and I try to install again, I receive this error, as if something has not be fully removed
```
4:22PM FTL /etc/systemd/system/uyuni-server.service file already present, not overwriting
```

cbosdo:
@mbussolotto was the installation you are removing successful or did it fail? That kind of behavior could happen if the service file was installed but the container isn't running

mbussolotto:
I can't remember, let's keep the issue opens so I remember to re-check it next time

mbussolotto:
fixed

PR #95, opened 2023-11-07
Fix errors messages in uyuniadm uninstall 

NamelessOne91:
Would help debugging issue #89


shared/utils/connection.go
```diff
@@ -31,16 +31,14 @@ func NewConnection(backend string) *Connection {
...
 	var err error
 	if c.command == "" {
-		command := ""
-
 		switch c.backend {
 		case "podman":
 			fallthrough
 		case "podman-remote":
-			fallthrough
+			c.command = c.backend
```
cbosdo, line 38: We want these two cases to check for the command to be in the `PATH`, so this change doesn't look OK

shared/utils/connection.go
```diff
@@ -31,16 +31,14 @@ func NewConnection(backend string) *Connection {
...
 		case "kubectl":
-			if _, err = exec.LookPath(command); err != nil {
-				err = fmt.Errorf("backend command not found in PATH: %s", command)
+			if _, err = exec.LookPath(c.backend); err != nil {
+				err = fmt.Errorf("backend command not found in PATH: %s", c.backend)
```
cbosdo, line 41: Here `c.backend` is usually not set: this code ensures we have the `podman` , `podman-remote` or `kubectl` command in the `PATH` before using it. So better remove this change

uyuniadm/cmd/uninstall/podman.go
```diff
@@ -21,9 +21,9 @@ func uninstallForPodman(dryRun bool, purge bool) {
 	} else {
 		if dryRun {
 			log.Info().Msgf("Would run systemctl disable --now uyuni-server")
-			log.Debug().Msgf("Woud remove %s", podman.ServicePath)
+			log.Debug().Msgf("Would remove %s", podman.ServicePath)
```
cbosdo, line 24: Thanks for spotting those typos!

uyuniadm/cmd/uninstall/podman.go
```diff
@@ -32,7 +32,9 @@ func uninstallForPodman(dryRun bool, purge bool) {
 
 			// Remove the service unit
 			log.Debug().Msgf("Remove %s", podman.ServicePath)
-			os.Remove(podman.ServicePath)
+			if err := os.Remove(podman.ServicePath); err != nil {
+				log.Error().Err(err).Msg("Failed to remove uyuni-server.service")
+			}
```
cbosdo, line 37: Thanks for adding that missing error check: this may help understanding the issue, but surely doesn't fix it.

cbosdo (changes requested):
Please drop the changes in the connection code as those would be leading to a lot of other issues. I am curious to know the reasons for those changes.

NamelessOne91:
Hi Cedric, thanks for the review.

I've dropped the changes to the connection-related code and kept only the additional error logs ATM.

As you stated in the issue's comments, the problem Michele had is likely due to not having a container named _uyuni-server_ running. This causes no backend to be picked up even if podman is present .
Is that working as intended ?

The lack of the `Msg` method call on line 28 of  [uninstall.go](https://github.com/uyuni-project/uyuni-tools/blob/main/uyuniadm/cmd/uninstall/uninstall.go)  may contribute to make this harder to spot.

Possible changes to the connection code are probably out of scope and may involve some refactoring, it might be better to discuss them as a separate topic.

cbosdo:
> As you stated in the issue's comments, the problem Michele had is likely due to not having a container named _uyuni-server_ running. This causes no backend to be picked up even if podman is present . Is that working as intended ?

No, this is not intended, but I haven't though very deeply how we should be handling that case. We probably need to try looking for other specific files and resources deployed early, like the systemd service file, the traefik config or similar.

> The lack of the `Msg` method call on line 28 of [uninstall.go](https://github.com/uyuni-project/uyuni-tools/blob/main/uyuniadm/cmd/uninstall/uninstall.go) may contribute to make this harder to spot.
> 
> Possible changes to the connection code are probably out of scope and may involve some refactoring, it might be better to discuss them as a separate topic.

Changing those is not easy and we should probably have a smarter detection only for uninstall in case nothing worked

cbosdo (approved):

cbosdo:
Could you please squash the two commits into one before I merge?

NamelessOne91:
Done, thanks.

cbosdo:
I'll postpone the merge until Uyuni 2023.10 is out

PR #96, opened 2023-11-08
Create commands for proxy

mbussolotto:


uyuniproxy/shared/templates/uyuni-proxy-httpd.go
```diff
@@ -0,0 +1,59 @@
...
+    --name {{ .NamePrefix }}-proxy-httpd ${UYUNI_IMAGE}
+
+ExecStop=/usr/bin/podman stop --ignore --cidfile %t/{{ .NamePrefix }}-proxy-httpd.ctr-id -t 10
+ExecStopPost=/usr/bin/podman rm --ignore -f --cidfile %t/{{ .NamePrefix }}-proxy-httpd.ctr-id
+PIDFile=%t/{{ .NamePrefix }}-proxy-httpd.pid
+TimeoutStopSec=60
+Type=forking
+
+[Install]
+WantedBy=multi-user.target default.target
```
aaannz, line 45: By the way, this type of running of podman is deprecated by podman upstream (see [Deprecated note](https://docs.podman.io/en/latest/markdown/podman-generate-systemd.1.html)).
They switched over to using systemd-generators and "Quadlet".

See https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html

Not sure if the SLE Micro 5.5 already have recent enough podman, also this way of starting podman containers will certainly work also in the future. So just that you know. :)

witekest: I found this discussion useful pointing the differences:

https://github.com/containers/podman/discussions/20305

_Note that there are no plans to remove podman-generate-systemd. Deprecation in this case means that podman-generate-systemd will not receive new features but only bug fixes. All new features go into Quadlet._

> Not sure if the SLE Micro 5.5 already have recent enough podman

Quadlet has been added in [Podman 4.4](https://github.com/containers/podman/blob/main/RELEASE_NOTES.md#440)  and I think is already included.

mbussolotto: Thanks for these comments! Although I was doing these test justs to play with Golang in a real scenario, from a quick glance seems that Quadlet might be something really helpful for us, maybe it allows us to share more code between podman and kubernetes. I'll have a deeper look on it !
@cbosdo

cbosdo: if quadlet is the future and has all we need, then go for it!

cbosdo: Let's forget about quadlet for now: it's not in Debian Stable and any released Ubuntu (only 24.04)

aaannz (approved):
LGTM

I am not certain if this is due to github interface, but navigating codebase in this PR was cognitively quite hard. For example there are `shared` and `utils` directory basically on every level so I need to keep in mind whole paths and frequently jump all over the place to find called code. Maybe having PR about refactoring, PR for podman and PR for k8s would be easier.

PR #97, opened 2023-11-08
Add -helm to default chart path to avoid conflict with server image

cbosdo:


deneb-alpha (approved):
LGTM.
For some context: without the  `-helm` the chart and   the container image will collide

Previous macro in the  prjconf of `systemsmanagement:Uyuni:Master:ContainerUtils` was:
~~~
# Definition of the default image and default chart to be used in case none is specified
Macros:
%_default_image registry.opensuse.org/systemsmanagement/uyuni/master/containers/uyuni/server
%_default_chart oci://registry.opensuse.org/systemsmanagement/uyuni/master/charts/uyuni/server
:Macros
~~~

PR #99, opened 2023-11-10
Hackweek: Quadlet server

mbussolotto:
Use Quadlet instead of systemd-generators 

TODO: uninstall


mbussolotto:
@cbosdo @aaannz the only great improvemente that I see it's that everything is more readable. I created the `network` file because probably it's something that user would like to modify. It's also possible to create `volume`  and `image` files (one for each volume and image), but I don't see great advance for now: probably it would be more useful when we'll have less volumes (or in proxy, where we have more than one image)

cbosdo:
> @cbosdo @aaannz the only great improvement that I see it's that everything is more readable. I created the `network` file because probably it's something that user would like to modify. It's also possible to create `volume` and `image` files (one for each volume and image), but I don't see great advance for now: probably it would be more useful when we'll have less volumes (or in proxy, where we have more than one image)

We don't want users to tamper with those generated files: it will only make it harder to update them if we need it. This is indeed way more readable for now, thanks for the work!

uyuniadm/shared/templates/containerTemplate.go
```diff
@@ -0,0 +1,53 @@
...
+	"text/template"
+
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
+)
+
+const containerTemplate = `
+[Unit]
+Description=Uyuni server image container service
+Wants=network.target
+After=network-online.target
```
cbosdo, line 14: We may want to add a dependency on mounts for the case where users mount external disks to the podman volumes path.

.gitignore
```diff
@@ -1,3 +1,4 @@
 bin
 .vscode/
 .idea/
+tags
```
cbosdo, line 4: I'm curious, what do you use it for? ctags?

mbussolotto: I think they're created by `ycm-core/YouCompleteMe` vim extension :) I configured it to autocomplete also go code

uyuniadm/shared/templates/networkTemplate.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"io"
+	"text/template"
+)
+
+const networkTemplate = `
+[Network]
+Options=
+NetworkName={{ .Network }}
```
rjmateus, line 11: Does this mean the code that creates the network should be adapted, or is independent?
https://github.com/uyuni-project/uyuni-tools/blob/main/uyuniadm/shared/podman/podman.go#L64-L112

mbussolotto: I need to verify it, but I think we can remove `setupNetwork`  and set the network here https://github.com/uyuni-project/uyuni-tools/blob/9041b2e77f680174f8945322995b12f16e3703b1/uyuniadm/shared/podman/podman.go#L62-L67

rjmateus: For that also consider the changes @cbosdo had to made to support ipv6

cbosdo:
I still wonder if this is a good thing: using quadlet will force a dependency on podman 4.4+. Which distros would that rule out as host?

cbosdo:
Distributions with podman 4.4+:
* Debian: trixie (testing)
* Ubuntu: Noble (24.04)
* Fedora: 37
* EL: 8
* Mageia: 8
* openSUSE Leap: 15.4
* SLE: 15 SP 4

cbosdo:
Closing for now: the benefit is not big enough to justify cutting support for Debian stable and Ubuntu

Issue #90, opened 2023-10-26
Solve debug info build errors


deneb-alpha:
Currently we disabled the debug build for
* [AlmaLinux_8](https://build.opensuse.org/project/repository_state/systemsmanagement:Uyuni:Master:ContainerUtils/AlmaLinux_8)
* [Fedora_38](https://build.opensuse.org/project/repository_state/systemsmanagement:Uyuni:Master:ContainerUtils/Fedora_38)

Due to `DWARF version 0 unhandled` failure. This is something already seen in the past when in SUSE Linux Enterprise we switched  to Go1.19 and it's something related  to go and elfutils/dwarves

See  also https://bugzilla.suse.com/show_bug.cgi?id=1203599. 
On SUSE Linux Enterprise we are fine but, at least for EL8 and Fedora 38 there's more work to be done

PR #105, opened 2023-11-20
Fix build for EL clones and Fedora

cbosdo:
Due to issues with the elf and go shipped in those distributions extracting debug symbols was failing. Not using compressed debug symbols works around the issue.

Along the same lines Fedora doesn't have the build-id, we need to force generating one.

Fixes https://github.com/uyuni-project/uyuni-tools/issues/90

### DIFF
--- a/uyuni-tools.spec
+++ b/uyuni-tools.spec
@@ -111,9 +111,21 @@ go_path=
   %endif
 %endif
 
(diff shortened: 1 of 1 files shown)


cbosdo:
Build fixed in my OBS branch with this change:
image

PR #107, opened 2023-11-23
Use git tag as version

cbosdo:
Instead of hardcoding the version, use the latest git tag and optionally an offset as version.


deneb-alpha (changes requested):
LGTM but, can we also have a changelog entry for this? :D

deneb-alpha:
@cbosdo one more "issue" I just notice. ATM you are editing directly the main .changes without adding any kind of header. 

I suppose we should have the tito property file added also to uyuni-tools. 
There's also the faketagger mechanism that could be helpful to have, but I'm not sure how that automation is currently enabled and done.

deneb-alpha:
> @cbosdo one more "issue" I just notice. ATM you are editing directly the main .changes without adding any kind of header.
> 
> I suppose we should have the tito property file added also to uyuni-tools. There's also the faketagger mechanism that could be helpful to have, but I'm not sure how that automation is currently enabled and done.

@juliogonzalez do you have more details about the faketagger part?

juliogonzalez:
> > @cbosdo one more "issue" I just notice. ATM you are editing directly the main .changes without adding any kind of header.
> > I suppose we should have the tito property file added also to uyuni-tools. There's also the faketagger mechanism that could be helpful to have, but I'm not sure how that automation is currently enabled and done.
> 
> @juliogonzalez do you have more details about the faketagger part (AKA how to enable it)?

The fake tagget is not part of tito, it's part of the `rel-eng` tooling in the main repository. When it detects the changelog is not "closed" (doesn't have a header), it adds the fake header before sending to OBS/IBS, but it does it only for the packages at the main repo.

Packages maintained outside the main repository do not have this, the changelog needs to be maintained and closed manually. Such is the case for salt-netapi-client, hub and others, and when packages want to submit a SR, they "close" the changelog entries, and submit.

Of course tito can be enabled, but for the fake tagger, you will need to implement a solution based on how this package is submitted to OBS/retrieved from OBS.

cbosdo:
> > > @cbosdo one more "issue" I just notice. ATM you are editing directly the main .changes without adding any kind of header.
> > > I suppose we should have the tito property file added also to uyuni-tools. There's also the faketagger mechanism that could be helpful to have, but I'm not sure how that automation is currently enabled and done.
> > 
> > 
> > @juliogonzalez do you have more details about the faketagger part (AKA how to enable it)?
> 
> The fake tagget is not part of tito, it's part of the `rel-eng` tooling in the main repository. When it detects the changelog is not "closed" (doesn't have a header), it adds the fake header before sending to OBS/IBS, but it does it only for the packages at the main repo.
> 
> Packages maintained outside the main repository do not have this, the changelog needs to be maintained and closed manually. Such is the case for salt-netapi-client, hub and others, and when packages want to submit a SR, they "close" the changelog entries, and submit.
> 
> Of course tito can be enabled, but for the fake tagger, you will need to implement a solution based on how this package is submitted to OBS/retrieved from OBS.

My idea is to move the `build-packages-for-obs.sh`, `push-packages-to-obs.sh` and the new `mkchlog` scripts to a separate repo for other repos to use and standardize how we work.

juliogonzalez:
> > The fake tagget is not part of tito, it's part of the `rel-eng` tooling in the main repository. When it detects the changelog is not "closed" (doesn't have a header), it adds the fake header before sending to OBS/IBS, but it does it only for the packages at the main repo.
> > Packages maintained outside the main repository do not have this, the changelog needs to be maintained and closed manually. Such is the case for salt-netapi-client, hub and others, and when packages want to submit a SR, they "close" the changelog entries, and submit.
> > Of course tito can be enabled, but for the fake tagger, you will need to implement a solution based on how this package is submitted to OBS/retrieved from OBS.
> 
> My idea is to move the `build-packages-for-obs.sh`, `push-packages-to-obs.sh` and the new `mkchlog` scripts to a separate repo for other repos to use and standardize how we work.

Could be a good idea, just keep in mind moving is not enough as refactors will be need, both on the code, and the Jenkins jobs.

cbosdo:
> Could be a good idea, just keep in mind moving is not enough as refactors will be need, both on the code, and the Jenkins jobs.

For a first step I will just duplicate the tools in a repo. Once it works for uyuni-tools then we can think about adapting to get rid of the one in `uyuni` repo

cbosdo:
Now it is possible to use `tito` for uyuni-tools. Use the tools from https://github.com/cbosdo/uyuni-releng-tools for it.

.gitignore
```diff
@@ -1,3 +1,4 @@
 bin
 .vscode/
 .idea/
+vendor.tar.gz
```
deneb-alpha, line 4: I think pycache should  be ignored too, right?
```suggestion
vendor.tar.gz
.tito/custom/__pycache__
```

uyuni-tools.changes
```diff
@@ -1,3 +1,4 @@
+- Use the latest git tag as version instead of hardcoding it
```
deneb-alpha, line 1: not sure if this is a leftover or what but I can see changes in the main .changes and also in the  custom one

deneb-alpha (changes requested):
two more comments  from my side. 

I see  the pycache uploaded (I suggested  to add it to gitignore) and also a new edit on  both .changes and your custom changes file

uyuni-tools.changes
cbosdo: I did this one before the changes. I can extract it too

deneb-alpha (approved):
LGTM, thanks a lot!

PR #109, opened 2023-11-27
Make REUSE compatible

cbosdo:
Make it compatible with https://reuse.software.
This removes the push-2-obs script: when merging this PR the automation needs to be adjusted


push-2-obs
deneb-alpha, line 1: sorry for the stupid question but I don't get why you have to remove the push-2-obs script

uyuni-tools.spec
deneb-alpha, line 1: spec-cleaner will re-add the header. I'm wondering if there are  any  best practices from pack-team

.vimrc
deneb-alpha, line 1: commenting on this but the remark applies to all the files where you applied CC0.

I know that REUSE suggests to apply CC0 to this kind of files, but CC0 is a license that when applied can't be revoked anymore because the author is waving all the rights. 

I'm  not 100% sure that SUSE legal will be happy with CC0 licensed files, even if they  are just config files.

deneb-alpha (commented):
left some comments

push-2-obs
cbosdo: This was a hack to duplicate the `push-packages-to-obs.sh` script

uyuni-tools.spec
cbosdo: I can add that one as config too to be on the safe side

.vimrc
cbosdo: Toggled all CC0 instances to Apache-2.0 but the reuse workflow that comes from FSF

PR #111, opened 2023-11-28
Add migration of autoinstallable distributions

nadvornik:
This PR implements copying of autoinstallable distributions to standard location `/srv/www/distributions`

Then the database must be updated with:
```
UPDATE rhnKickstartableTree SET base_path = CONCAT('/srv/www/distributions/', CASE WHEN org_id = 1 THEN label ELSE CONCAT(org_id, '-', label) END);
```
This is not part of the PR.

### DIFF
--- a/mgradm/shared/templates/migrateScriptTemplate.go
+++ b/mgradm/shared/templates/migrateScriptTemplate.go
@@ -17,15 +17,42 @@ if test -e /tmp/ssh_config; then
 fi
 SSH="ssh -A $SSH_CONFIG "
 SCP="scp -A $SSH_CONFIG "
+
+echo "Stopping spacewalk service..."
+$SSH {{ .SourceFqdn }} "spacewalk-service stop ; systemctl start postgresql.service"
+
+$SSH {{ .SourceFqdn }} \
+ "echo \"COPY (SELECT MIN(CONCAT(org_id, '-', label)) AS target, base_path FROM rhnKickstartableTree GROUP BY base_path) TO STDOUT WITH CSV;\" \
+ |spacewalk-sql --select-mode - " > distros
+
+echo "Stopping posgresql service..."
+$SSH {{ .SourceFqdn }} "systemctl stop postgresql.service"
+
+while IFS="," read -r target path ; do
+    echo "-/ $path"
+done < distros > exclude_list
+
 for folder in {{ range .Volumes }}{{ . }} {{ end }};
 do
   if $SSH {{ .SourceFqdn }} test -e $folder; then
     echo "Copying $folder..."
-    rsync -e "$SSH" --rsync-path='sudo rsync' -avz {{ .SourceFqdn }}:$folder/ $folder;
+    rsync -e "$SSH" --rsync-path='sudo rsync' -avz -f "merge exclude_list" {{ .SourceFqdn }}:$folder/ $folder;
   else
     echo "Skipping missing $folder..."
   fi
 done;
+
+echo "Migrating auto-installable distributions..."
+while IFS="," read -r target path ; do
+  if $SSH -A {{ .SourceFqdn }} test -e $path; then
+    echo "Copying distribution $target from $path"
+    mkdir -p "/srv/www/distributions/$target"
+    rsync -e "$SSH" --rsync-path='sudo rsync' -avz "{{ .SourceFqdn }}:$path/" "/srv/www/distributions/$target"
+  else
+    echo "Skipping missing distribution $path..."
+  fi
+done < distros
+
 rm -f /srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT;
 ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT;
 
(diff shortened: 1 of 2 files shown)


uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -13,15 +13,35 @@ if test -e /tmp/ssh_config; then
 fi
 SSH="ssh -A $SSH_CONFIG "
 SCP="scp -A $SSH_CONFIG "
+
+ssh -A {{ .SourceFqdn }} \
```
srbarrios, line 17: `$SSH` maybe?

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -13,15 +13,35 @@ if test -e /tmp/ssh_config; then
 fi
 SSH="ssh -A $SSH_CONFIG "
 SCP="scp -A $SSH_CONFIG "
+
```
srbarrios, line 16: I think it would be good to have an echo before this part starts, something like...

echo "Migrating auto-installable distributions"

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -13,15 +13,35 @@ if test -e /tmp/ssh_config; then
...
+ssh -A {{ .SourceFqdn }} \
+ "echo \"COPY (SELECT CASE WHEN org_id = 1 THEN label ELSE CONCAT(org_id, '-', label) END AS target, base_path FROM rhnKickstartableTree) TO STDOUT WITH CSV;\" \
```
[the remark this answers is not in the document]
nadvornik: I think it will need `spacewalk-service stop` and `systemctl start posgresql` at the beginning of the script
and  `systemctl stop posgresql`  after this line.

Regarding org_id, the idea is that for org id's different from 1 the name will be prefixed by the id. This should not cause any conflicts.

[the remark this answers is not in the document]
rjmateus: @nadvornik why don't we keep it consistent and always add the org ID and prefix to the label, and this way we can have the same behavior.

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -13,15 +13,35 @@ if test -e /tmp/ssh_config; then
...
   else
     echo "Skipping missing $folder..."
   fi
 done;
+
+while IFS="," read -r target path ; do
+  if ssh -A {{ .SourceFqdn }} test -e $path; then
+    echo "Copying distribution $target from $path"
+    mkdir -p "/srv/www/distributions/$target"
+    rsync -e "$SSH" --rsync-path='sudo rsync' -avz "{{ .SourceFqdn }}:$path/" "/srv/www/distributions/$target"
```
rjmateus, line 39: The value in the `$targe` will be "org+label" of the database entry.
If it got this right, in case the same distribution folder is used by two different `rhnKickstartableTree` table entries in different organizations, we will be duplicating it.
In some special cases it can duplicate the distributions and disk space in use.

I was thinking if we could look at the last folder name in the "$path" and use it, but that can lead to collisions.

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -13,15 +13,35 @@ if test -e /tmp/ssh_config; then
...
+  else
+    echo "Skipping missing distribution $path..."
+  fi
+done < distros
```
rjmateus, line 43: I think we are missing the update clause in the Database to cope with the new location.

nadvornik:
I have updated it according to comments. Now it should handle also 2 distributions in the same or different org, pointing to the same directory.

The database must be updated with this, but it is not clear yet when to do it:
```
SELECT MIN(CONCAT(id, '-', label)) AS target, base_path INTO TEMP TABLE dist_map FROM rhnKickstartableTree GROUP BY base_path;
UPDATE rhnKickstartableTree SET base_path = CONCAT('/srv/www/distributions/', target) from dist_map where dist_map.base_path = rhnKickstartableTree.base_path;
drop table dist_map;

```

cbosdo (approved):

uyuniadm/shared/templates/migrateScriptTemplate.go
cbosdo: > I think we are missing the update clause in the Database to cope with the new location.

Vladimir stated in the PR description this is not in this PR.

mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -17,15 +17,42 @@ if test -e /tmp/ssh_config; then
 fi
 SSH="ssh -A $SSH_CONFIG "
 SCP="scp -A $SSH_CONFIG "
+
+echo "Stopping spacewalk service..."
+$SSH {{ .SourceFqdn }} "spacewalk-service stop ; systemctl start postgresql.service"
+
+$SSH {{ .SourceFqdn }} \
+ "echo \"COPY (SELECT MIN(CONCAT(id, '-', label)) AS target, base_path FROM rhnKickstartableTree GROUP BY base_path) TO STDOUT WITH CSV;\" \
```
rjmateus, line 25: I'm not sure about this query. What will be the behaviour if the same path is used in two different label/id. Since the values will be different and are not part of the group by, I don't know how it will behave.
One more thing, the ID is the row identifier. I think you want to use the field `org_id`

nadvornik:
I have changed `id` to `org_id`.

The database update is here:
https://github.com/uyuni-project/uyuni-tools/pull/116

rjmateus (approved):

PR #112, opened 2023-11-29
Use the SSH variable instead of ssh command in migrate script

srbarrios:
In order to load the ssh_config file, we want to use `$SSH` instead of `ssh`


srbarrios:
Merging, this is a minimal change that works.

PR #114, opened 2023-11-30
Migration script: Altering configuration for domain resolution

srbarrios:
**Context:**
Once we migrate from one server to another, we have a problem with domain resolution, as we have all the configurations using the FQDN from the previous machine.

**Manual step after migration:**
Ideally, we expect our users to re-use the same FQDN in the new machine once the migration script has finished.

**Proposal:**

Here we replace some of the FQDN matches in our etc config files, routing them to `localhost`, the changed values must stay for the correct usage on containers.
**But** we will still have some configuration files using the FQDN of the source server, therefore we still require that the user manually replace the FQDN of the new server and re-use the FQDN of the source server.

### DIFF
--- a/uyuniadm/shared/templates/migrateScriptTemplate.go
+++ b/uyuniadm/shared/templates/migrateScriptTemplate.go
@@ -32,6 +32,11 @@ ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/
 echo "Extracting time zone..."
 $SSH {{ .SourceFqdn }} timedatectl show -p Timezone >/var/lib/uyuni-tools/data
 
+echo "Altering configuration for domain resolution..."
+sed 's/report_db_host = {{ .SourceFqdn }}/report_db_host = localhost/' -i /etc/rhn/rhn.conf;
+sed 's/server\.jabber_server/java\.hostname/' -i /etc/rhn/rhn.conf;
+sed 's/client_use_localhost: false/client_use_localhost: true/' -i /etc/cobbler/settings.yaml;
+
 {{ if .Kubernetes }}
 echo "Altering configuration for kubernetes..."
 echo 'server.no_ssl = 1' >> /etc/rhn/rhn.conf;


cbosdo:
> **Proposal:** So, I'm not convinced of this draft, but as I have been discussing this idea today with Cedric, and it allows me to test on my VMs without hacks for resolution, I'm open this draft to discuss with all of you.
> 
> Here we replace the FQDN matches in our etc config files, routing them to `localhost`. This could be a temporary state where we will allow the user to test the status of the system, but once they replace the FQDN of the new machine, they will need to replace those `localhost` matches again.

I just remembered that some of the values are used to compute URLs to be used by the clients or the UI and should not be changed. Those who are used by the server only can (and should) stay as `localhost` as this will avoid possible network hair pins with the podman network.

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -32,6 +32,11 @@ ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/
 echo "Extracting time zone..."
 $SSH {{ .SourceFqdn }} timedatectl show -p Timezone >/var/lib/uyuni-tools/data
 
+echo "Altering configuration for domain resolution..."
+sed 's/= {{ .SourceFqdn }}/= localhost/' -i /etc/rhn/rhn.conf;
```
cbosdo, line 36: In `rhn.conf`, only the value of `report_db_host` needs to be set to `localhost`.
* `server.jabber_server`  should be kept and converted to `java.hostname`
* `cobber.host` should be kept as is
* the others (osad and jabber) are unused in 5.0

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -32,6 +32,11 @@ ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/
...
+sed 's/{{ .SourceFqdn }}/localhost/' -i /etc/cobbler/settings.yaml;
```
cbosdo, line 37: My bad, in cobbler's settings the change is not the FQDN, but `client_use_localhost` to be toggled from `false` to `true`

uyuniadm/shared/templates/migrateScriptTemplate.go
```diff
@@ -32,6 +32,11 @@ ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/
...
+sed 's/{{ .SourceFqdn }}/localhost/' -i /etc/salt/master.d/uyuni_roster.conf
```
cbosdo, line 38: We are not setting this value to `localhost` in the setup of a new container: this should not be in the migration script.

cbosdo (changes requested):
There is still a point that is not clear: how to we convert the 4.3 config files to match those from 5.0? How did we do this in the past major versions bumps?

uyuniadm/shared/templates/migrateScriptTemplate.go
srbarrios: Ok, perfect I will do these changes then, including the conversion of the property `java.hostname`.

uyuniadm/shared/templates/migrateScriptTemplate.go
srbarrios: Cool, an easy one to fix too, I will change it.

uyuniadm/shared/templates/migrateScriptTemplate.go
srbarrios: Done.

uyuniadm/shared/templates/migrateScriptTemplate.go
srbarrios: Done

uyuniadm/shared/templates/migrateScriptTemplate.go
srbarrios: Pablo confirmed we can't change this value.
We can only wait until the user replaces the FQDN in the VM.

cbosdo (approved):

PR #116, opened 2023-12-12
Add postgres migration

nadvornik:
This PR adds the logic for database migration, including distro migration (followup of https://github.com/uyuni-project/uyuni-tools/pull/111 )

### DIFF
--- a/mgradm/cmd/migrate/kubernetes/utils.go
+++ b/mgradm/cmd/migrate/kubernetes/utils.go
@@ -8,18 +8,19 @@ package kubernetes
 
 import (
 	"encoding/base64"
+	"fmt"
 	"os"
 	"path"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
-	"github.com/uyuni-project/uyuni-tools/shared/types"
-	"github.com/uyuni-project/uyuni-tools/shared/utils"
 	"github.com/uyuni-project/uyuni-tools/mgradm/cmd/migrate/shared"
 	"github.com/uyuni-project/uyuni-tools/mgradm/shared/kubernetes"
 	"github.com/uyuni-project/uyuni-tools/mgradm/shared/ssl"
 	adm_utils "github.com/uyuni-project/uyuni-tools/mgradm/shared/utils"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
 func migrateToKubernetes(globalFlags *types.GlobalFlags, flags *kubernetesMigrateFlags, cmd *cobra.Command, args []string) {
@@ -40,6 +41,7 @@ func migrateToKubernetes(globalFlags *types.GlobalFlags, flags *kubernetesMigrat
 	// We don't need the SSL certs at this point of the migration
 	clusterInfos := kubernetes.CheckCluster()
 
+	kubeconfig := clusterInfos.GetKubeconfig()
 	//TODO: check if we need to handle SELinux policies, as we do in podman
 
 	kubernetes.Deploy(cnx, &flags.Image, &flags.Helm, &sslFlags, &clusterInfos, fqdn, false,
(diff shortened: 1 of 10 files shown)


mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -83,15 +87,112 @@ rm -rf /root/ssl-build
 # The content of this folder will be a RO mount from a configmap
 rm /etc/pki/trust/anchors/*
 {{ end }}
+
+echo "DONE"`
+
+const podmanPostgresVersionMigrationScriptTemplate = `#!/bin/bash
```
rjmateus, line 93: Instead of generating this script in here, would make sense to re-use and adapt the existing infrastructure we have in here: https://github.com/uyuni-project/uyuni/blob/master/susemanager/bin/pg-migrate-x-to-y.sh

We would need to add the migration script for the right versions and install the package `susemanager` in the migration container

nadvornik: there are several problems with the script as it is now:
- we need 2 parts, each run in different container
- the script starts postgres and other services via systemd, this does not work in the container
- it needs susemanager package, as you mentioned

I think that separate script for containers is easier to maintain.

rjmateus: because of all the thinks you mention, I agree with you. :)

mgradm/shared/utils/cmd_utils.go
```diff
@@ -87,3 +92,8 @@ func AddImageFlag(cmd *cobra.Command) {
 	cmd.Flags().String("pullPolicy", "IfNotPresent",
 		"set whether to pull the images or not. The value can be one of 'Never', 'IfNotPresent' or 'Always'")
 }
+
+func AddMigrationImageFlag(cmd *cobra.Command) {
+	cmd.Flags().String("migrationImage", "", "Migration Image")
+	cmd.Flags().String("migrationImageTag", DefaultTag, "Migration Image Tag")
```
cbosdo, line 98: May be it would be better with `migration-image` and `migration-tag`. This would make the configuration in the yaml file like:

```
migration:
  image: the/migration/image
  tag: 1234
```

mgradm/shared/utils/cmd_utils.go
```diff
@@ -36,6 +36,11 @@ type ImageFlags struct {
 	PullPolicy string `mapstructure:"pullPolicy"`
 }
 
+type MigrationImageFlags struct {
+	Name       string `mapstructure:"migrationImage"`
+	Tag        string `mapstructure:"migrationImageTag"`
+}
+
```
cbosdo, line 43: You can use the `ImageFlags` structure above even if `PullPolicy` is not used in the case. And with the parameters changes suggested below you only have to define `mapstructure:"migration"` on the field where this is used.

mgradm/cmd/migrate/shared/flags.go
```diff
@@ -11,8 +11,10 @@ import (
 
 type MigrateFlags struct {
 	Image cmd_utils.ImageFlags `mapstructure:",squash"`
+	MigrationImage cmd_utils.MigrationImageFlags `mapstructure:",squash"`
```
cbosdo, line 14: With the parameters change suggested below you only need this:
```suggestion
	MigrationImage cmd_utils.MigrationImageFlags `mapstructure:"migration"`
```

cbosdo (commented):
It looks really good. I have some cosmetic suggestions you may want to look into

mgradm/cmd/migrate/kubernetes/utils.go
```diff
@@ -51,14 +53,31 @@ func migrateToKubernetes(globalFlags *types.GlobalFlags, flags *kubernetesMigrat
...
+	tz, oldPgVersion, newPgVersion := shared.ReadContainerData(scriptDir)
 
 	helmArgs := []string{
 		"--reset-values",
 		"--set", "timezone=" + tz,
 	}
 
-	kubeconfig := clusterInfos.GetKubeconfig()
+	if oldPgVersion != newPgVersion {
+		migrationImage := flags.Image
```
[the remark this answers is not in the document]
nadvornik: The idea is that the migration image will be located next to server image, so it should share the info, unless specified otherwise.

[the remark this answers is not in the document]
cbosdo: The `flags.Image` value is the full path to the server image, not the folder where it is contained.

PR #120, opened 2024-01-09
Look for the helm release when uninstalling the server

cbosdo:
If the release install failed, there is no pod but we still need to uninstall the heln release or the next install will fail.


mbussolotto (approved):
LGTM. There's a conflict in the file but looks an easy one

PR #121, opened 2024-01-12
Add migration of config files

nadvornik:
With this PR the migration command copies from the old machine only files marked config(noreplace) in rpm.
Additionally, it completely excludes db migration and makes sure that the migrated files points to /usr/share/susemanager.

Fixes https://github.com/SUSE/spacewalk/issues/23259

### DIFF
--- a/mgradm/shared/templates/migrateScriptTemplate.go
+++ b/mgradm/shared/templates/migrateScriptTemplate.go
@@ -32,6 +32,18 @@ while IFS="," read -r target path ; do
     echo "-/ $path"
 done < distros > exclude_list
 
+# exclude all config files which already exist and are not marked noreplace
+rpm -qa --qf '[%{fileflags},%{filenames}\n]' |grep ",/etc/" | while IFS="," read -r flags path ; do
+    # config(noreplace) is 1<<4 (from lib/rpmlib.h)
+    if [ $(( $flags & 16 )) -eq 0 -a -f "$path" ] ; then
+        echo "-/ $path" >> exclude_list
+    fi
+done
+
+# exclude schema migration files
+echo "-/ /etc/sysconfig/rhn/reportdb-schema-upgrade" >> exclude_list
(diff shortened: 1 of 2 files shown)


mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -32,6 +32,18 @@ while IFS="," read -r target path ; do
     echo "-/ $path"
 done < distros > exclude_list
 
+# exclude all config files which already exist and are not marked noreplace
+rpm -qa --qf '[%{fileflags},%{filenames}\n]' |grep ",/etc/" | while IFS="," read -r flags path ; do
```
cbosdo, line 36: Shouldn't we run this `rpm` command over SSH on the source machine? I believe it should give identical results, but that could be a source of errors.

mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -32,6 +32,18 @@ while IFS="," read -r target path ; do
...
+    # config(noreplace) is 1<<4 (from lib/rpmlib.h)
+    if [ $(( $flags & 16 )) -eq 0 -a -f "$path" ] ; then
+        echo "-/ $path" >> exclude_list
+    fi
+done
+
+# exclude schema migration files
+echo "-/ /etc/sysconfig/rhn/reportdb-schema-upgrade" >> exclude_list
+echo "-/ /etc/sysconfig/rhn/schema-upgrade" >> exclude_list
```
cbosdo, line 45: Nice! and this way the script doesn't need to be touched fast after the schema upgrade files are moved away.

mgradm/shared/templates/migrateScriptTemplate.go
nadvornik: I think we want to look at the newer RPMs from the container. This way we can fix eventual errors in the old RPMs.

cbosdo (approved):

PR #122, opened 2024-01-15
Fix the build tags in the spec file

cbosdo:
The default image, tag and chart have been refactored with the addition of `mgrpxy` but the spec file hasn't been adjusted to it yet.

The project configuration in OBS will now need a definition like:

    Macros:
    %_default_namespace registry.suse.de/devel/galaxy/manager/head/containers/suse/manager/5.0
    %_default_tag 5.0.0-alpha1
    :Macros

The `%_default_image` and `%_default_chart` variables should be dropped as they are no longer used.

Since the default values are hard coded for Uyuni, there is no OBS project configuration to change for it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,8 +35,9 @@ jobs:
 
       - name: Compute version
         run: |
-          tag=$(git describe --tags --abbrev=0 | cut -f 3 -d '-')
-          offset=$(git rev-list --count ${tag})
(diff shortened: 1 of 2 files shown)


juliogonzalez (approved):

raulillo82 (approved):

deneb-alpha (approved):
it  seems   ok to me

PR #124, opened 2024-01-17
Remove custom SELinux policy, instead use label:disable for podman migration

aaannz:
Fixes SELinux access problems for SSH agent socket by disabling selinux relabeling for migration container.

This does not affect container host, side effect is that for migration container there is no separation in selinux labels, weakening a security by little. However migration container is short lived one and once migration is complete, normal server container with separate selinux labels for container is used as before.

I did not test Kubernetes yet, but I expect it to have similar problem with accessing SSH agent and will need similar solution by using something like:

```yaml
apiVersion: v1
kind: Pod
spec:
  containers:
    securityContext:
        seLinuxOptions:
            type: spc_t

```

Fixes https://github.com/SUSE/spacewalk/issues/23125

### DIFF
--- a/mgradm/cmd/migrate/podman/utils.go
+++ b/mgradm/cmd/migrate/podman/utils.go
@@ -28,6 +28,7 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *podmanMigrateFlags, 
 	defer os.RemoveAll(scriptDir)
 
 	extraArgs := []string{
+		"--security-opt", "label:disable",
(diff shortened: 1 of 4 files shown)


aaannz:
CC @srbarrios

srbarrios (approved):
Awesome!
If we managed to do it without changing SELinux policies much better!

I wonder if our salt coleagues might follow the same strategy for their case 🤔

cbosdo (approved):

PR #125, opened 2024-01-17
Add mgradm support config command

cbosdo:
Run the support config on the host, in the container and pack them into a `tar.gz` for the user to submit to the support team.


cbosdo:
Fixes https://github.com/SUSE/spacewalk/issues/22109

shared/utils/tar.go
```diff
@@ -70,3 +71,61 @@ func ExtractTarGz(tarballPath string, dstPath string) error {
...
+// Object holding a .tar.gz to write it to a file.
+type TarGz struct {
+	fileWriter *os.File
+	tarWriter  *tar.Writer
+	gzipWriter *gzip.Writer
+}
+
+// NewTarGz create a targz object with writers opened.
+// A successful call should be followed with a close.
+func NewTarGz(path string) (*TarGz, error) {
```
rjmateus, line 84: Instead of creating this object to creat a tar, would be better to call a cmd to compress the wanted file with the `tar` command instead?

shared/utils/tar.go
```diff
@@ -0,0 +1,72 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/rs/zerolog/log"
+)
+
+// Extracts a tar.gz file.
+func ExtractTarGz(tarballPath string, dstPath string) error {
```
rjmateus, line 20: the same comment as before, would make sense to just call `tar` command instead of implementing like this?

rjmateus (commented):
Looks good to me. Just 2 question about the tar usage, if we should delegate to the `tar` command instead of implementing it with go

shared/utils/tar.go
cbosdo: Now that it works and is unit tested what would be the benefit? Using a command for such a thing could make the whole thing even harder to unit test (which I haven't done yet)

shared/utils/tar.go
cbosdo: Same answer here

rjmateus (approved):

mbussolotto (approved):
LGTM

PR #129, opened 2024-01-19
Accept image with the tag in --image parameters

cbosdo:
Many users are used to append the tag to the image, accept this as a valid value for --image. In such a case any value of --tag will be ignored.


mbussolotto (approved):
LGTM

PR #130, opened 2024-01-19
Add push script to tweak spec file for IBS

cbosdo:
There are some build differences between the tools for Uyuni and those for SUSE Manager. Add a push.sh script to apply those changes when pushing to IBS for SUSE Manager.

Requires PR https://github.com/uyuni-project/uyuni-releng-tools/pull/9 to be merged and deployed before pushing.

### DIFF
--- a/.tito/custom/custom.py
+++ b/.tito/custom/custom.py
@@ -9,7 +9,7 @@ Code for building packages in SUSE that need generated code not tracked in git.
 import os
 
 from tito.builder import Builder
-from tito.common import  info_out, run_command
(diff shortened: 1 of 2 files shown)


deneb-alpha (commented):
@cbosdo probably a changelog could be helpful too  given that we introduce the new push.sh here, what's your take?

Apart from this, the rest LGTM

cbosdo:
> @cbosdo probably a changelog could be helpful too given that we introduce the new push.sh here, what's your take?
> 
> Apart from this, the rest LGTM

It's not of interest for customers, so I'ld avoid it.

PR #135, opened 2024-01-22
Use a TestMain func for utils tests setup/teardown

NamelessOne91:
This PR is a proposal (and example of) aimed at moving the setup/teardown functions needed by the utils package tests to a dedicated file.
Using`testing.M` it will be possible to have a single setup function, running before  the package's tests, and a single teardown function , running after them.

This should make the setup/teardown of temporary directories and files needed by the tests in that package less prone to repetition as the testsuite grows.

### DIFF
--- a/shared/utils/tar_test.go
+++ b/shared/utils/tar_test.go
@@ -11,111 +11,58 @@ import (
 	"testing"
 )
 
-const dataDir = "data"
-const outDir = "out"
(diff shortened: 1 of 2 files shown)


cbosdo:
Thanks for the hint, but the tar tests should be quite isolated: we don't need the data files for other tests.

PR #141, opened 2024-01-31
stop gracefully postgresql and spacewalk when stop uyuni-server container

mbussolotto:
stop gracefully postgresql and spacewalk when stop uyuni-server container


cbosdo (changes requested):
Please invert the `spacewalk-service stop` and `systemctl stop postgresql`

cbosdo (approved):

PR #142, opened 2024-01-31
store UYUNI IMAGE values in a separated config file

mbussolotto:
store UYUNI IMAGE values in a separated config file

### DIFF
--- a/mgradm/shared/templates/serviceTemplate.go
+++ b/mgradm/shared/templates/serviceTemplate.go
@@ -22,7 +22,6 @@ RequiresMountsFor=%t/containers
 
 [Service]
 Environment=PODMAN_SYSTEMD_UNIT=%n
-Environment=UYUNI_IMAGE={{ .Image }}
 Environment=TZ={{ .Timezone }}
 Restart=on-failure
 ExecStartPre=/bin/rm -f %t/uyuni-server.pid %t/%n.ctr-id
--- a/shared/podman/systemd.go
+++ b/shared/podman/systemd.go
@@ -78,11 +78,26 @@ func IsServiceRunning(service string) bool {
 	return cmd.ProcessState.ExitCode() == 0
 }
 
+// CmdService send a command to the systemd service.
+func CmdService(service string, cmd string) {
+	if err := utils.RunCmd("systemctl", cmd, service); err != nil {
+		log.Fatal().Err(err).Msgf("Failed to %s systemd %s.service", cmd, service)
+	}
+}
+
 // RestartService restarts the systemd service.
 func RestartService(service string) {
-	if err := utils.RunCmd("systemctl", "restart", service); err != nil {
-		log.Fatal().Err(err).Msgf("Failed to restart systemd %s.service", service)
-	}
+	CmdService(service, "restart")
+}
+
+// StopService stop the systemd service.
+func StopService(service string) {
+	CmdService(service, "stop")
+}
+
+// StopService stop the systemd service.
+func StartService(service string) {
+	CmdService(service, "start")
 }
 
 // EnableService enables and starts a systemd service.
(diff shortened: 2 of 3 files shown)


rjmateus (approved):

mgradm/shared/templates/serviceTemplate.go
```diff
@@ -22,7 +22,6 @@ RequiresMountsFor=%t/containers
 
 [Service]
 Environment=PODMAN_SYSTEMD_UNIT=%n
-Environment=UYUNI_IMAGE={{ .Image }}
```
cbosdo, line 25: You don't have to remove it from here. If there is an override file with that value podman will use it instead. This is what we currently do with the proxy. This way you don't need to create the override config file for the first install, only updates.

shared/podman/systemd.go
```diff
@@ -78,11 +78,26 @@ func IsServiceRunning(service string) bool {
 	return cmd.ProcessState.ExitCode() == 0
 }
 
+// CmdService send a command to the systemd service.
+func CmdService(service string, cmd string) {
+	if err := utils.RunCmd("systemctl", cmd, service); err != nil {
+		log.Fatal().Err(err).Msgf("Failed to %s systemd %s.service", cmd, service)
+	}
+}
```
cbosdo, line 86:
```suggestion
func CmdService(service string, cmd string) error {
	if err := utils.RunCmd("systemctl", cmd, service); err != nil {
		return fmt.Errorf("Failed to %s systemd %s.service: %s", cmd, service, err)
	}
	return nil
}
```

shared/podman/systemd.go
```diff
@@ -78,11 +78,26 @@ func IsServiceRunning(service string) bool {
...
+
 // RestartService restarts the systemd service.
 func RestartService(service string) {
-	if err := utils.RunCmd("systemctl", "restart", service); err != nil {
-		log.Fatal().Err(err).Msgf("Failed to restart systemd %s.service", service)
-	}
+	CmdService(service, "restart")
```
cbosdo, line 90:
```suggestion
	return CmdService(service, "restart")
```

shared/podman/systemd.go
```diff
@@ -78,11 +78,26 @@ func IsServiceRunning(service string) bool {
...
+}
+
+// StopService stop the systemd service.
+func StopService(service string) {
+	CmdService(service, "stop")
+}
```
cbosdo, line 96:
```suggestion
func StopService(service string) error {
	return CmdService(service, "stop")
}
```

shared/podman/systemd.go
```diff
@@ -78,11 +78,26 @@ func IsServiceRunning(service string) bool {
...
+
+// StopService stop the systemd service.
+func StartService(service string) {
+	CmdService(service, "start")
```
cbosdo, line 100:
```suggestion
func StartService(service string) error {
	return CmdService(service, "start")
```

shared/podman/systemd.go
```diff
@@ -91,3 +106,30 @@ func EnableService(service string) {
 		log.Fatal().Err(err).Msgf("Failed to enable %s systemd service", service)
 	}
 }
+
+// Create new conf file
+func GenerateSystemdConfFile(serviceName string, section string, body string) {
```
cbosdo, line 111:
```suggestion
func GenerateSystemdConfFile(serviceName string, section string, body string) error {
```

shared/podman/systemd.go
```diff
@@ -91,3 +106,30 @@ func EnableService(service string) {
...
+	systemdFilePath := GetServicePath(serviceName)
+	log.Info().Msgf("systemdFilePath: %s", systemdFilePath)
+
+	systemdConfFolder := systemdFilePath + ".d"
+	log.Info().Msgf("systemdConfFolder: %s", systemdConfFolder)
+	if err := os.MkdirAll(systemdConfFolder, 0750); err != nil {
+		log.Fatal().Err(err).Msgf("Failed to create %s folder", systemdConfFolder)
```
cbosdo, line 118:
```suggestion
		return fmt.Errorf("Failed to create %s folder: %s", systemdConfFolder, err)
```

shared/podman/systemd.go
```diff
@@ -91,3 +106,30 @@ func EnableService(service string) {
...
+	log.Info().Msgf("systemdConfFilePath: %s", systemdConfFilePath)
+
+	if utils.FileExists(systemdConfFilePath) {
+		log.Warn().Msgf("File %s already exists. It will be override", systemdConfFilePath)
+	}
+
+	content := []byte("[" + section + "]" + "\n" + body + "\n")
+
+	if err := os.WriteFile(systemdConfFilePath, content, 0644); err != nil {
+		log.Fatal().Err(err).Msgf("Cannot write %s file", systemdConfFilePath)
```
cbosdo, line 132:
```suggestion
		fmt.Errorf("Cannot write %s file: %s", systemdConfFilePath, err)
```

shared/podman/systemd.go
```diff
@@ -91,3 +106,30 @@ func EnableService(service string) {
...
+	}
```
cbosdo, line 133:
```suggestion
	}
	return nil
```

cbosdo (changes requested):
error handling suggestions

cbosdo (changes requested):

shared/podman/systemd.go
```diff
@@ -78,11 +79,27 @@ func IsServiceRunning(service string) bool {
...
 }
 
-// RestartService restarts the systemd service.
-func RestartService(service string) {
-	if err := utils.RunCmd("systemctl", "restart", service); err != nil {
-		log.Fatal().Err(err).Msgf("Failed to restart systemd %s.service", service)
+// CmdService send a command to the systemd service.
+func CmdService(service string, cmd string) error {
+	if err := utils.RunCmd("systemctl", cmd, service); err != nil {
+		return fmt.Errorf("Failed to %s systemd %s.service: %s", cmd, service, err)
```
cbosdo, line 85: This won't fly with localization: the `cmd` will have to stay in english for `systemctl` to understand it but we also need it localized in the error message. I'm not convinced we need to factorize those.

PR #148, opened 2024-02-05
Add first user creation to `mgradm install` and `mgrctl`

cbosdo:
While at it I started structuring a bit more the code for API calls.


aaannz (approved):
👍🏾

PR #150, opened 2024-02-07
Add generated code for API in mgrctl

cbosdo:
The code has been generated using a Javadoc doclet from the API documentation. The code used for this is available in the following PR:

https://github.com/uyuni-project/uyuni/pull/8272


cbosdo:
Probably not what we want. Closing for now

PR #151, opened 2024-02-07
force system list refresh on migrate

rjmateus:
Insert directly in the queue to update all the systems in the system list view in case we don't have run of the overall update task.
I have tested it locally in a migration from 4.3.10 to HEAD running in podman.

### DIFF
--- a/mgradm/shared/templates/pgsqlPostMigrationScriptTemplate.go
+++ b/mgradm/shared/templates/pgsqlPostMigrationScriptTemplate.go
@@ -38,6 +38,17 @@ UPDATE rhnKickstartableTree SET base_path = CONCAT('/srv/www/distributions/', ta
 DROP TABLE dist_map;
 EOT
 {{ end }}
+
+echo "Schedule a system list update task..."
+spacewalk-sql --select-mode - <<EOT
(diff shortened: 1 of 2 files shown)


mgradm/shared/templates/pgsqlPostMigrationScriptTemplate.go
```diff
@@ -38,6 +38,17 @@ UPDATE rhnKickstartableTree SET base_path = CONCAT('/srv/www/distributions/', ta
 DROP TABLE dist_map;
 EOT
 {{ end }}
+
+echo "Schedule a system list update task..."
+spacewalk-sql --select-mode - <<EOT
+insert into rhnTaskQueue (id, org_id, task_name, task_data)
+SELECT nextval('rhn_task_queue_id_seq'), 1, 'update_system_overview', s.id
+from rhnserver s
```
cbosdo, line 46: This will only for update for systems of Organization 1. We surely want to loop on the organizations to refresh them all.

cbosdo (changes requested):

nadvornik (approved):
LGTM

PR #170, opened 2024-02-22
ignore error on optional flag

mbussolotto:
ignore error on optional flag. This should allow to continue compile `mgrctl`


mcalmer (approved):
I do not understand go, but sounds ok

PR #193, opened 2024-03-06
Add data checks to user input

cbosdo:
## What does this PR change?

Add length and content check to user input. Constraints extracted from https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/action/multiorg/validation/orgCreateForm.xsd#L25

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23337

- [X] **DONE**

## Changelogs

- [X] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/install/shared/flags.go
+++ b/mgradm/cmd/install/shared/flags.go
@@ -5,6 +5,11 @@
 package shared
 
 import (
+	"fmt"
+	"net/mail"
+	"regexp"
+	"strings"
+
 	"github.com/spf13/cobra"
 	cmd_utils "github.com/uyuni-project/uyuni-tools/mgradm/shared/utils"
 	apiTypes "github.com/uyuni-project/uyuni-tools/shared/api/types"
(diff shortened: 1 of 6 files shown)


mgradm/cmd/install/shared/flags.go
```diff
@@ -56,30 +60,50 @@ type InstallFlags struct {
...
 }
 
+// idChecker verifies that the value is a valid identifier.
+func idChecker(value string) bool {
+	isAscii := regexp.MustCompile(`^[[:graph:]]+$`)
+	if isAscii.MatchString(value) && !strings.ContainsAny(value, "+&'\"`=#") {
+		return true
+	}
+	fmt.Println("Can only contain ASCII characters without +&'\"`=#")
+	return false
```
aaannz, line 70: So do we allow e.g. `>` in the id?

Wouldn't `[[:alnum:]]` be better here?

mgradm/cmd/install/shared/flags.go
```diff
@@ -56,30 +60,50 @@ type InstallFlags struct {
...
+}
+
+// emailChecker verifies that the value is a valid email address.
+func emailChecker(value string) bool {
+	r := regexp.MustCompile(`^(.+)@(.+)(\..+)*$`)
+	if !r.MatchString(value) {
+		fmt.Println("Not a valid email address")
+		return false
+	}
+	return true
```
aaannz, line 80: This is quite naive, I understand it's taken from Java part, but we have https://pkg.go.dev/net/mail#ParseAddress

shared/utils/utils.go
```diff
@@ -19,15 +19,41 @@ import (
...
+		tmpValue := strings.TrimSpace(string(bytePassword))
+		r := regexp.MustCompile(`^[^\t ]+$`)
+		validChars := r.MatchString(tmpValue)
+		if !validChars {
+			fmt.Printf("Cannot contain spaces or tabs")
+		}
+
+		if validChars && checkValueSize(tmpValue, min, max) {
+			*value = tmpValue
+		}
```
aaannz, line 56: I am sorry, but this is some go-fu which I am not familiar with.
What does `for *value == "" {` should do? How can this function fail if the format is not valid?

mgradm/cmd/install/shared/flags.go
cbosdo: well, in the Java code it's even worse, we also include control characters 😉 But yes, I could just change to `[[:alnum:]]`

shared/utils/utils.go
cbosdo: This is like `for ; *value == "" ; {`. This only stops with a correct value or a `Ctrl-C`.

PR #206, opened 2024-03-19
allow installation/upgrade/ migration using RPM image

mbussolotto:
## What does this PR change?

allow installation/upgrade/ migration using RPM image

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23682

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/install/kubernetes/kubernetes.go
+++ b/mgradm/cmd/install/kubernetes/kubernetes.go
@@ -16,6 +16,7 @@ import (
 
 type kubernetesInstallFlags struct {
 	shared.InstallFlags `mapstructure:",squash"`
+	Image               types.ImageFlags `mapstructure:",squash"`
 	Helm                cmd_utils.HelmFlags
 }
 
--- a/mgradm/cmd/install/kubernetes/kubernetes.go
+++ b/mgradm/cmd/install/kubernetes/kubernetes.go
@@ -16,7 +16,6 @@ import (
 
 type kubernetesInstallFlags struct {
 	shared.InstallFlags `mapstructure:",squash"`
-	Image               types.ImageFlags `mapstructure:",squash"`
 	Helm                cmd_utils.HelmFlags
 }
 
--- a/shared/podman/images.go
+++ b/shared/podman/images.go
@@ -19,74 +19,79 @@ import (
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
+const rpmImageDir = "/usr/share/suse-docker-images/native/"
+
+var registries = []string{
+	"registry.suse.com",
+	"registry.opensuse.com",
+}
+
 // Ensure the container image is pulled or pull it if the pull policy allows it.
+//
+// Returns the image name to use. Note that it may be changed if the image has been loaded from a local RPM package.
 func PrepareImage(image string, pullPolicy string, args ...string) (string, error) {
-	log.Info().Msgf("Ensure image %s is available", image)
+	if strings.ToLower(pullPolicy) != "always" {
+		log.Info().Msgf("Ensure image %s is available", image)
 
-	present, err := IsImagePresent(image)
-	if err != nil {
-		return "", err
+		presentImage, err := IsImagePresent(image)
+		if err != nil {
+			return image, err
+		}
+
+		if len(presentImage) > 0 {
+			log.Debug().Msgf("Image %s already present", presentImage)
+			return presentImage, nil
+		}
 	}
 
-	if present && strings.ToLower(pullPolicy) != "always" {
-		log.Debug().Msgf("Image %s already present", image)
-		return "", nil
-	}
-
-	rpmImageFile, err := IsRpmImagePresent(image)
-	if err != nil {
-		return "", err
-	}
+	rpmImageFile := GetRpmImagePath(image)
 
 	if len(rpmImageFile) > 0 {
 		log.Debug().Msgf("Image %s present as RPM. Loading it", image)
 		loadedImage, err := loadRpmImage(rpmImageFile)
 		if err != nil {
-			log.Warn().Msgf("cannot use RPM image for %s:%s", image, err)
-			present = false
+			log.Warn().Msgf("Cannot use RPM image for %s: %s", image, err)
 		} else {
-			if strings.ToLower(pullPolicy) == "always" {
-				log.Debug().Msg("Ignoring pull policy alway ")
-			}
-
-			log.Warn().Msgf("Loading image %s: it's the RPM based image of %s.", strings.TrimSpace(loadedImage), image)
+			log.Info().Msgf("Using the %s image loaded from the RPM instead of its online version %s", strings.TrimSpace(loadedImage), image)
 			return loadedImage, nil
 		}
+	} else {
+		log.Info().Msgf("Cannot find RPM image for %s", image)
 	}
 
-	if strings.ToLower(pullPolicy) == "always" {
-		log.Debug().Msgf("Pulling image cause pull policy is always %s", image)
+	if strings.ToLower(pullPolicy) != "never" {
+		log.Debug().Msgf("Pulling image %s because it is missing and pull policy is not 'never'", image)
 		return image, pullImage(image, args...)
 	}
 
-	if !present && strings.ToLower(pullPolicy) != "never" {
-		log.Debug().Msgf("Pulling image cause is missing and pull policy is not never %s", image)
-		return image, pullImage(image, args...)
-	}
-
-	return image, fmt.Errorf("image %s is missing and cannot be fetch", image)
+	return image, fmt.Errorf("image %s is missing and cannot be fetched", image)
 }
 
-// GetRpmInfoFromImage return the RPM Image name and the tag, given an image.
-func GetRpmInfoFromImage(image string) (rpmImageFile string, tag string) {
-	rpmImageFile = strings.ReplaceAll(image, "registry.suse.com/", "")
-	rpmImageFile = strings.ReplaceAll(rpmImageFile, "/", "-")
-	parts := strings.Split(rpmImageFile, ":")
-	tag = "latest"
-	if len(parts) > 1 {
-		tag = parts[1]
+// GetRpmImageName return the RPM Image name and the tag, given an image.
+func GetRpmImageName(image string) (rpmImageFile string, tag string) {
+	for _, registry := range registries {
+		if strings.HasPrefix(image, registry) {
+			rpmImageFile = strings.ReplaceAll(image, registry+"/", "")
+			rpmImageFile = strings.ReplaceAll(rpmImageFile, "/", "-")
+			parts := strings.Split(rpmImageFile, ":")
+			tag = "latest"
+			if len(parts) > 1 {
+				tag = parts[1]
+			}
+			rpmImageFile = parts[0]
+			return rpmImageFile, tag
+		}
 	}
-	rpmImageFile = parts[0]
-	return rpmImageFile, tag
+	return "", ""
 }
 
-// GetRpmInfoFromImage return the path of an RPM Image.
-func GetRpmImage(byteValue []byte, rpmImageFile string, tag string) (string, error) {
+// BuildRpmImagePath checks the image metadata and returns the RPM Image path.
+func BuildRpmImagePath(byteValue []byte, rpmImageFile string, tag string) (string, error) {
 	var data types.Metadata
 	if err := json.Unmarshal(byteValue, &data); err != nil {
 		return "", fmt.Errorf("cannot unmarshal: %s", err)
 	}
-	fullPathFile := "/usr/share/suse-docker-images/native/" + data.Image.File
+	fullPathFile := rpmImageDir + data.Image.File
 	if data.Image.Name == rpmImageFile {
 		for _, metadataTag := range data.Image.Tags {
 			if metadataTag == tag {
@@ -97,16 +102,16 @@ func GetRpmImage(byteValue []byte, rpmImageFile string, tag string) (string, err
 	return "", nil
 }
 
-// IsRpmImagePresent return true if the RPM with the provided image is installed.
-func IsRpmImagePresent(image string) (string, error) {
-	log.Debug().Msgf("looking for RPM image for %s", image)
+// GetRpmImagePath return the RPM image path.
+func GetRpmImagePath(image string) string {
+	log.Debug().Msgf("Looking for installed RPM package containing %s image", image)
 
-	rpmImageFile, tag := GetRpmInfoFromImage(image)
+	rpmImageFile, tag := GetRpmImageName(image)
 
-	rpmImageDir := "/usr/share/suse-docker-images/native/"
 	files, err := os.ReadDir(rpmImageDir)
 	if err != nil {
-		return "", fmt.Errorf("cannot read directory %s: %s", rpmImageDir, err)
+		log.Debug().Msgf("Cannot read directory %s: %s", rpmImageDir, err)
+		return ""
 	}
 
 	for _, file := range files {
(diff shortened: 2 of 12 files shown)


mgradm/cmd/install/kubernetes/kubernetes.go
```diff
@@ -16,6 +16,7 @@ import (
 
 type kubernetesInstallFlags struct {
 	shared.InstallFlags `mapstructure:",squash"`
+	Image               types.ImageFlags `mapstructure:",squash"`
```
cbosdo, line 19: This feels useless: this is not the way to define the image to use on kubernetes

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+	"path"
 	"strings"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
 // Ensure the container image is pulled or pull it if the pull policy allows it.
```
cbosdo, line 22:
```suggestion
// Ensure the container image is pulled or pull it if the pull policy allows it.
//
// Returns the image name to use. Note that it may be changed if the image has been loaded from a local RPM package.
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+		loadedImage, err := loadRpmImage(rpmImageFile)
+		if err != nil {
+			log.Warn().Msgf("cannot use RPM image for %s:%s", image, err)
+			present = false
+		} else {
+			if strings.ToLower(pullPolicy) == "always" {
+				log.Debug().Msg("Ignoring pull policy alway ")
+			}
+
+			log.Warn().Msgf("Loading image %s: it's the RPM based image of %s.", strings.TrimSpace(loadedImage), image)
```
cbosdo, line 52:
```suggestion
			log.Warn().Msgf("Using the %s image loaded from the RPM instead of its online version %s", strings.TrimSpace(loadedImage), image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+		log.Debug().Msgf("Pulling image cause pull policy is always %s", image)
+		return image, pullImage(image, args...)
+	}
+
+	if !present && strings.ToLower(pullPolicy) != "never" {
+		log.Debug().Msgf("Pulling image cause is missing and pull policy is not never %s", image)
+		return image, pullImage(image, args...)
+	}
+
+	return image, fmt.Errorf("image %s is missing and cannot be fetch", image)
```
cbosdo, line 67:
```suggestion
	return image, fmt.Errorf("image %s is missing and cannot be fetched", image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+			return loadedImage, nil
+		}
 	}
-	return nil
-}
 
-func checkImage(image string, pullPolicy string) (bool, error) {
 	if strings.ToLower(pullPolicy) == "always" {
-		return true, nil
+		log.Debug().Msgf("Pulling image cause pull policy is always %s", image)
```
cbosdo, line 58:
```suggestion
		log.Debug().Msgf("Pulling image %s because pull policy is 'always'", image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+	rpmImageFile, err := GetRpmImagePath(image)
+	if err != nil {
+		return image, err
+	}
+
+	if len(rpmImageFile) > 0 {
+		log.Debug().Msgf("Image %s present as RPM. Loading it", image)
+		loadedImage, err := loadRpmImage(rpmImageFile)
+		if err != nil {
+			log.Warn().Msgf("cannot use RPM image for %s:%s", image, err)
```
cbosdo, line 45:
```suggestion
			log.Warn().Msgf("Cannot use RPM image for %s: %s", image, err)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+			present = false
+		} else {
+			if strings.ToLower(pullPolicy) == "always" {
+				log.Debug().Msg("Ignoring pull policy alway ")
```
cbosdo, line 49:
```suggestion
				log.Debug().Msg("Ignoring pull policy 'always'")
```
Should we really ignore the pull policy? I would just apply it: if the image is already loaded, just skip.

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
 
-func checkImage(image string, pullPolicy string) (bool, error) {
 	if strings.ToLower(pullPolicy) == "always" {
-		return true, nil
+		log.Debug().Msgf("Pulling image cause pull policy is always %s", image)
+		return image, pullImage(image, args...)
+	}
+
+	if !present && strings.ToLower(pullPolicy) != "never" {
+		log.Debug().Msgf("Pulling image cause is missing and pull policy is not never %s", image)
```
cbosdo, line 63:
```suggestion
		log.Debug().Msgf("Pulling image %s because it is missing and pull policy is not 'never'", image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+				return fullPathFile, nil
+			}
+		}
+	}
+	return "", nil
+}
+
+// GetRpmImagePath return the RPM image path.
+func GetRpmImagePath(image string) (string, error) {
+	log.Debug().Msgf("looking for RPM image for %s", image)
```
cbosdo, line 102:
```suggestion
	log.Debug().Msgf("Looking for installed RPM package containing %s image", image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+
+	rpmImageFile, tag := GetRpmImageName(image)
+
+	rpmImageDir := "/usr/share/suse-docker-images/native/"
```
cbosdo, line 106: Move this to a constant in the module to reuse it. It will make it easier to change if it has to change in the future.

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+	if err != nil {
+		return "", fmt.Errorf("cannot read directory %s: %s", rpmImageDir, err)
+	}
+
+	for _, file := range files {
+		if !strings.HasSuffix(file.Name(), "metadata") {
+			continue
+		}
+		fullPathFileName := path.Join(rpmImageDir, file.Name())
+		log.Debug().Msgf("parsing %s", fullPathFileName)
```
cbosdo, line 117:
```suggestion
		log.Debug().Msgf("Parsing metadata file %s", fullPathFileName)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+		fileHandler, err := os.Open(fullPathFileName)
+		if err != nil {
+			log.Debug().Msgf("error opening %s: %s", fullPathFileName, err)
```
cbosdo, line 120:
```suggestion
			log.Debug().Msgf("Error opening metadat file %s: %s", fullPathFileName, err)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+			continue
+		}
+		defer fileHandler.Close()
+		byteValue, err := io.ReadAll(fileHandler)
+		if err != nil {
+			log.Debug().Msgf("error reading %s: %s", fullPathFileName, err)
```
cbosdo, line 126:
```suggestion
			log.Debug().Msgf("Error reading metadata file %s: %s", fullPathFileName, err)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+			continue
+		}
+
+		fullPathFile, err := BuildRpmImagePath(byteValue, rpmImageFile, tag)
+		if err != nil {
+			log.Warn().Msgf("cannot unmarshal %s: %s", fullPathFileName, err)
```
cbosdo, line 132:
```suggestion
			log.Warn().Msgf("Cannot unmarshal metadata file %s: %s", fullPathFileName, err)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
+			return "", err
+		}
+		if len(fullPathFile) > 0 {
+			log.Debug().Msgf("%s match with %s", fullPathFileName, image)
+			return fullPathFile, nil
+		}
+		log.Debug().Msgf("%s does not match with %s", fullPathFileName, image)
+	}
+	log.Debug().Msgf("image %s does not exists as RPM image", image)
```
cbosdo, line 141:
```suggestion
	log.Debug().Msgf("No installed RPM package containing %s image", image)
```

shared/podman/images.go
```diff
@@ -6,48 +6,175 @@ package podman
...
-	if needsPull {
-		return pullImage(image, args...)
+	if present && strings.ToLower(pullPolicy) != "always" {
+		log.Debug().Msgf("Image %s already present", image)
+		return image, nil
+	}
+
+	rpmImageFile, err := GetRpmImagePath(image)
+	if err != nil {
+		return image, err
```
cbosdo, line 38: This returns an error if there is no installed RPM image... is that what we want if we the image should be pulled from a registry?

mbussolotto: right, the intention was to raise an error but not catch it. I'll fix it.

shared/podman/images.go
mbussolotto: ok, I was not sure about the best to do.

shared/podman/images.go
cbosdo: What's the point in raising an error if you don't catch it?

shared/podman/images.go
cbosdo: I wouldn't ignore the policy: that would be too confusing for users

shared/podman/images.go
mbussolotto: Because it's not an error here inside `PrepareImage`, but it can be an error used somewhere else. From a function called `GetRpmImagePath` I'm expecting that raise an error in the image is not find. But if you think it's useless I can remove it

cbosdo: that's fine... but the wording was strange

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+	"path"
 	"strings"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
+var rpmImageDir = "/usr/share/suse-docker-images/native/"
```
cbosdo, line 22:
```suggestion
const rpmImageDir = "/usr/share/suse-docker-images/native/"
```

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
-		return err
+		return image, err
+	}
+
+	if present {
+		log.Debug().Msgf("Image %s already present", image)
+		return image, nil
+	}
+
+	rpmImageFile, _ := GetRpmImagePath(image)
```
cbosdo, line 45: I don't like the idea to totally ignore an error: this can hide real errors

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
-func PrepareImage(image string, pullPolicy string, args ...string) error {
+//
+// Returns the image name to use. Note that it may be changed if the image has been loaded from a local RPM package.
+func PrepareImage(image string, pullPolicy string, args ...string) (string, error) {
 	log.Info().Msgf("Ensure image %s is available", image)
 
-	needsPull, err := checkImage(image, pullPolicy)
+	if strings.ToLower(pullPolicy) == "always" {
+		log.Debug().Msgf("Pulling image %s because pull policy is 'always'", image)
+		return image, pullImage(image, args...)
```
cbosdo, line 32: This means that if pull policy is `always` the RPM image will not be loaded? Maybe you should keep `checkImage()` and move the RPM image discovery at this level. This way you would know:
* the real image name to look for
* if the image has been loaded already

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+			return "", err
+		}
+		if len(fullPathFile) > 0 {
+			log.Debug().Msgf("%s match with %s", fullPathFileName, image)
+			return fullPathFile, nil
+		}
+		log.Debug().Msgf("%s does not match with %s", fullPathFileName, image)
+	}
+	log.Debug().Msgf("No installed RPM package containing %s image", image)
+	return "", fmt.Errorf("image %s does not exists as RPM image", image)
```
cbosdo, line 137:
```suggestion
	return "", nil
```
The fact that the file can't be found is not an error in itself: you don't want to break the install because of it. You are already using a check of the file path length to figure out if the RPM has been found: raising an error would have it at the same level than errors to look for the file itself.

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+		byteValue, err := io.ReadAll(fileHandler)
+		if err != nil {
+			log.Debug().Msgf("Error reading metadata file %s: %s", fullPathFileName, err)
+			continue
+		}
+
+		fullPathFile, err := BuildRpmImagePath(byteValue, rpmImageFile, tag)
+		if err != nil {
+			log.Warn().Msgf("Cannot unmarshal metadata file %s: %s", fullPathFileName, err)
+			return "", err
```
cbosdo, line 128: This is a valid error for this function, but I start to wonder if that error should return errors at all: shouldn't it just add error messages and let the install try with online images?

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+
+// GetRpmImagePath return the RPM image path.
+func GetRpmImagePath(image string) (string, error) {
+	log.Debug().Msgf("Looking for installed RPM package containing %s image", image)
+
+	rpmImageFile, tag := GetRpmImageName(image)
 
+	files, err := os.ReadDir(rpmImageDir)
+	if err != nil {
+		return "", fmt.Errorf("cannot read directory %s: %s", rpmImageDir, err)
```
cbosdo, line 104: If no RPM is installed, this will fail and we probably don't want such a behavior

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+
+	return image, fmt.Errorf("image %s is missing and cannot be fetched", image)
 }
 
-func checkImage(image string, pullPolicy string) (bool, error) {
-	if strings.ToLower(pullPolicy) == "always" {
-		return true, nil
+// GetRpmImageName return the RPM Image name and the tag, given an image.
+func GetRpmImageName(image string) (rpmImageFile string, tag string) {
+	rpmImageFile = strings.ReplaceAll(image, "registry.suse.com/", "")
```
cbosdo, line 68: We may want to handle the images on registry.opensuse.org at some point too for offline installations.

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+		log.Debug().Msgf("%s does not match with %s", fullPathFileName, image)
+	}
+	log.Debug().Msgf("No installed RPM package containing %s image", image)
+	return "", fmt.Errorf("image %s does not exists as RPM image", image)
+}
+
+func loadRpmImage(rpmImageBasePath string) (string, error) {
+	out, err := utils.RunCmdOutput(zerolog.DebugLevel, "podman", "load", "--quiet", "--input", rpmImageBasePath)
+	if err != nil {
+		return "", fmt.Errorf("cannot load image from: %s, continuing trying to pull from the registry: %s", rpmImageBasePath, err)
```
cbosdo, line 143: This message will be concatenated with another and look like this:
```
Cannot use RPM image for %s: cannot load image from: %s, continuing trying to pull from the registry: %s
```
I would just return the error here:
```suggestion
		return "", err
```

shared/podman/images.go
```diff
@@ -6,48 +6,170 @@ package podman
...
+	}
+	loadedImage := strings.ReplaceAll(string(out), "Loaded image: ", "")
```
cbosdo, line 145: I hope `podman` won't localize its output or this would break

mbussolotto: safer to `split` and take the last value :)

shared/podman/images.go
mbussolotto: right, looks like errors here are just generating confusion. Better to remove them at all

shared/podman/images.go
mbussolotto: removed it

shared/podman/images.go
mbussolotto: with latest code, `always` does not check if the image is present: it re-load RPM image (if present) and then pull the image

cbosdo (approved):

PR #253, opened 2024-04-18
add uyuni attestation container for podman

lucidd:
## What does this PR change?

add support for coco attestation container to mgradm.

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/23556

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge


mgradm/cmd/install/podman/utils.go
```diff
@@ -23,6 +23,28 @@ import (
...
 
+func setupCocoContainer(flags *podmanInstallFlags) error {
+	if flags.Coco.Containers > 0 {
+		tag := flags.Coco.Image.Tag
+		if tag == "" {
+			tag = flags.Image.Tag
+		}
+		cocoImage, err := utils.ComputeImage(flags.Coco.Image.Name, tag)
+		if err != nil {
+			return fmt.Errorf("failed to compute image URL, %s", err)
```
mbussolotto, line 34: Now that we have localization, all these errors needs to be inside `L` function (e.g. https://github.com/uyuni-project/uyuni-tools/blob/68cf416f14bd878c58bdafadd7fcc4c1657e7b1e/mgradm/shared/podman/podman.go#L66 . You need to import `. "github.com/uyuni-project/uyuni-tools/shared/l10n"` (e.g. https://github.com/uyuni-project/uyuni-tools/blob/68cf416f14bd878c58bdafadd7fcc4c1657e7b1e/mgradm/shared/podman/podman.go#L21C2-L21C54 )

mgradm/cmd/upgrade/podman/utils.go
```diff
@@ -64,6 +68,10 @@ func upgradePodman(globalFlags *types.GlobalFlags, flags *podmanUpgradeFlags, cm
 		return fmt.Errorf(L("cannot run post upgrade script: %s"), err)
 	}
 
+	if err := shared_podman.GenerateSystemdConfFile("uyuni-server-attestation", "Service", "Environment=UYUNI_IMAGE="+serverImage); err != nil {
```
mbussolotto, line 71: you can re-use `shared_podman.ServerAttestationService` instead of `"uyuni-server-attestation"`

mgradm/shared/podman/podman.go
```diff
@@ -47,6 +47,29 @@ func GetExposedPorts(debug bool) []types.PortMap {
...
+		Image:      image,
+	}
+	if err := utils.WriteTemplateToFile(attestationData, podman.GetServicePath(podman.ServerAttestationService), 0555, false); err != nil {
+		return fmt.Errorf("failed to generate systemd service unit file: %s", err)
+	}
+
+	environment := fmt.Sprintf(`Environment=UYUNI_IMAGE=%s
+Environment=database_connection=jdbc:postgresql://uyuni-server:5432/susemanager
+Environment=database_user=spacewalk
+Environment=database_password=spacewalk
```
mbussolotto, line 64: if `database_connection`, `database_user` and `database_password` would not change, I'd put them in `attestationServiceTemplate`. The idea is to have the main service file with values that doesn't change and the folder with values that might change (e.g. `UYUNI_IMAGE`)

mbussolotto (commented):
I added some comment, the rest looks fine (but let's wait for other review :) )

mgradm/cmd/install/podman/utils.go
mbussolotto: (this is the reason why this test is failing https://github.com/uyuni-project/uyuni-tools/actions/runs/8739643458/job/23981418583?pr=253 )

mgradm/shared/podman/podman.go
mcalmer: I think these values should change. What is the idea how to do this?

mbussolotto: if they might change, this is ok. We might improve trying to have separate `conf` file for different type of variable.
We this change we would have `Service.conf` that contains:
```
Environment=UYUNI_IMAGE=THE_IMAGE_URL
Environment=database_connection=jdbc:postgresql://uyuni-server:5432/susemanager
Environment=database_user=spacewalk
Environment=database_password=spacewalk
```
Probably the best can be have an `image.conf` with :+1: 
```
Environment=UYUNI_IMAGE=THE_IMAGE_URL
```
and a `database.conf` with:
```
Environment=database_connection=jdbc:postgresql://uyuni-server:5432/susemanager
Environment=database_user=spacewalk
Environment=database_password=spacewalk
```
as you prefer, this approach is good as well.

lucidd: those values can be changed as they refer to the uyuni db. Another reason to have it all here in one place is that the service file is world readable while this dropin config directory is not.

mgradm/shared/templates/attestationServiceTemplate.go
```diff
@@ -0,0 +1,62 @@
...
+	--cidfile=%t/%n.ctr-id \
+	--cgroups=no-conmon \
+	--sdnotify=conmon \
+	-d \
+	-e database_connection  \
+	-e database_user \
+	-e database_password \
+	--replace \
+	--name {{ .NamePrefix }}-server-attestation \
+	--hostname {{ .NamePrefix }}-server-attestation \
```
mcalmer, line 36: I think we should use the same internal domain as the server-image has.
```suggestion
	--hostname {{ .NamePrefix }}-server-attestation.mgr.internal \
```

mgradm/shared/podman/podman.go
```diff
@@ -47,6 +48,29 @@ func GetExposedPorts(debug bool) []types.PortMap {
...
+		NamePrefix: "uyuni",
+		Network:    podman.UyuniNetwork,
+		Image:      image,
+	}
+	if err := utils.WriteTemplateToFile(attestationData, podman.GetServicePath(podman.ServerAttestationService), 0555, false); err != nil {
+		return fmt.Errorf(L("failed to generate systemd service unit file: %s"), err)
+	}
+
+	environment := fmt.Sprintf(`Environment=UYUNI_IMAGE=%s
+Environment=database_connection=jdbc:postgresql://uyuni-server:%d/%s
```
mcalmer, line 63:
```suggestion
Environment=database_connection=jdbc:postgresql://uyuni-server.mgr.internal:%d/%s
```

We use now a domain for the internal network.

mcalmer (changes requested):
I added a domain for the internal network. I think we should use this everywhere.
I wonder if it works without, but maybe yes, as it than only use the short hostname.

mgradm/cmd/restart/podman.go
```diff
@@ -16,5 +16,9 @@ func podmanRestart(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	return podman.RestartService(podman.ServerService)
+	err := podman.RestartService(podman.ServerService)
+	if err != nil {
+		return err
+	}
+	return podman.RestartService(podman.ServerAttestationService)
```
cbosdo, line 23: Wouldn't that fail if the number of replicas is set to 0?  I would check if the service is defined first

mgradm/cmd/start/podman.go
```diff
@@ -16,5 +16,9 @@ func podmanStart(
 	cmd *cobra.Command,
 	args []string,
 ) error {
+	err := podman.StartService(podman.ServerAttestationService)
```
cbosdo, line 19: Same here... wouldn't that fail if the number of replicas is set to 0? I would check if the service is defined first

mgradm/cmd/stop/podman.go
```diff
@@ -16,5 +16,9 @@ func podmanStop(
 	cmd *cobra.Command,
 	args []string,
 ) error {
+	err := podman.StopService(podman.ServerAttestationService)
```
cbosdo, line 19: Same here... wouldn't that fail if the number of replicas is set to 0? I would check if the service is defined first

mgradm/cmd/uninstall/podman.go
```diff
@@ -23,9 +23,11 @@ func uninstallForPodman(
 ) error {
 	// Uninstall the service
 	podman.UninstallService("uyuni-server", !flags.Force)
+	podman.UninstallService(podman.ServerAttestationService, !flags.Force)
```
cbosdo, line 26: Same here... wouldn't that fail if the number of replicas is set to 0? I would check if the service is defined first

mgradm/cmd/uninstall/podman.go
```diff
@@ -23,9 +23,11 @@ func uninstallForPodman(
...
 
 	// Force stop the pod
 	podman.DeleteContainer(podman.ServerContainerName, !flags.Force)
+	podman.DeleteContainer(podman.ServerAttestationService, !flags.Force)
```
cbosdo, line 30: Same here... wouldn't that fail if the number of replicas is set to 0? I would check if the service is defined first

mgradm/cmd/upgrade/podman/utils.go
```diff
@@ -38,6 +38,10 @@ func upgradePodman(globalFlags *types.GlobalFlags, flags *podmanUpgradeFlags, cm
 		return err
 	}
 
+	if err := shared_podman.StopService(shared_podman.ServerAttestationService); err != nil {
+		return fmt.Errorf(L("cannot stop service %s"), err)
+	}
```
cbosdo, line 43: Same here... wouldn't that fail if the number of replicas is set to 0? I would check if the service is defined first

mgradm/shared/templates/attestationServiceTemplate.go
```diff
@@ -0,0 +1,62 @@
...
+ExecStop=/usr/bin/podman stop --ignore -t 10 --cidfile=%t/%n.ctr-id
+ExecStopPost=/usr/bin/podman rm -f --ignore -t 10 --cidfile=%t/%n.ctr-id
+PIDFile=%t/uyuni-server-attestation.pid
+TimeoutStopSec=60
+TimeoutStartSec=60
+Type=forking
+
+[Install]
+WantedBy=multi-user.target default.target
+`
```
cbosdo, line 49: You may want to create a systemd unit template as documented in [systemd.unit(5)](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#):

>Unit names can be parameterized by a single argument called the "instance name". The unit is then constructed based on a "template file" which serves as the definition of multiple services or other units. A template unit must have a single "@" at the end of the unit name prefix (right before the type suffix). The name of the full unit is formed by inserting the instance name between "@" and the unit type suffix. In the unit file itself, the instance parameter may be referred to using "%i" and other specifiers, see below.

This would allow you to spawn multiple instances of the same coco container to get multiple replicas.

The template could look like this:

```suggestion
const attestationServiceTemplate = `# uyuni-server-attestation.service, generated by mgradm
# Use an uyuni-server-attestation.service.d/local.conf file to override

[Unit]
Description=Uyuni server attestation container service
Wants=network.target
After=network-online.target

[Service]
Environment=PODMAN_SYSTEMD_UNIT=%n
Restart=on-failure
ExecStartPre=/bin/rm -f %t/uyuni-server-attestation-%i.pid %t/%n.ctr-id
ExecStartPre=/usr/bin/podman rm --ignore --force -t 10 {{ .NamePrefix }}-server-attestation-%i
ExecStart=/usr/bin/podman run \
	--conmon-pidfile %t/uyuni-server-attestation-%i.pid \
	--cidfile=%t/%n-%i.ctr-id \
	--cgroups=no-conmon \
	--sdnotify=conmon \
	-d \
	-e database_connection  \
	-e database_user \
	-e database_password \
	--replace \
	--name {{ .NamePrefix }}-server-attestation-%i \
	--hostname {{ .NamePrefix }}-server-attestation.mgr.internal \
	--network {{ .Network }} \
	${UYUNI_IMAGE}

ExecStop=/usr/bin/podman stop --ignore -t 10 --cidfile=%t/%n-%i.ctr-id
ExecStopPost=/usr/bin/podman rm -f --ignore -t 10 --cidfile=%t/%n-%i.ctr-id
PIDFile=%t/uyuni-server-attestation-%i.pid
TimeoutStopSec=60
TimeoutStartSec=60
Type=forking

[Install]
WantedBy=multi-user.target default.target
`
```

I'm not sure about the `-%i` variable. I'm also not sure if the hostname should include the variable and if the folder to override the properties would stay the same.

mgradm/shared/utils/cmd_utils.go
```diff
@@ -74,6 +74,14 @@ func AddHelmInstallFlag(cmd *cobra.Command) {
 	_ = utils.AddFlagToHelpGroupID(cmd, "helm-certmanager-values", "helm")
 }
 
+// AddContainerImageFlags add container image flags to command.
+func AddContainerImageFlags(cmd *cobra.Command, container string) {
+	cmd.Flags().String(container+"-image", "",
+		fmt.Sprintf("Image for %s container, overrides the namespace if set", container))
```
cbosdo, line 80:
```suggestion
		fmt.Sprintf(L("Image for %s container, overrides the namespace if set"), container))
```

mgradm/shared/utils/cmd_utils.go
```diff
@@ -74,6 +74,14 @@ func AddHelmInstallFlag(cmd *cobra.Command) {
...
+	cmd.Flags().String(container+"-tag", "",
+		fmt.Sprintf("Tag for %s container, overrides the global value if set", container))
```
cbosdo, line 82:
```suggestion
		fmt.Sprintf(L("Tag for %s container, overrides the global value if set"), container))
```

mgradm/shared/templates/attestationServiceTemplate.go
cbosdo: Fixed

mgradm/shared/podman/podman.go
cbosdo: Fixed

mcalmer (approved):
my concerns are addressed. I will leave the rest for the experts :-)

mgradm/shared/templates/attestationServiceTemplate.go
lucidd: the plan was to use systemd templates for this but since we don't want to add support for more then one container until later, i will leave this out for now.

PR #278, opened 2024-04-26
L10n context

cbosdo:
## What does this PR change?

* Add `PL()` function to provide context for translators.
* Remove the trailing `: %s` for errors from the localized strings.
* Ensure we don't have multiple unnamed placeholders in strings by adding a position flag.

## Test coverage

- No tests: added a Github action to detect the multiple unnamed placeholders.

- [X] **DONE**

## Links

Issue(s): #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge


shared/api/api.go
```diff
@@ -109,7 +109,7 @@ func (c *HTTPClient) sendRequest(req *http.Request) (*http.Response, error) {
 
 	res, err := c.Client.Do(req)
 	if err != nil {
-		log.Trace().Msgf("Request failed: %s", err)
+		log.Trace().Err(err).Msgf("Request failed")
```
mbussolotto, line 112: Missing localization.

shared/podman/images.go
```diff
@@ -111,7 +112,7 @@ func GetRpmImagePath(image string) string {
 
 	files, err := os.ReadDir(rpmImageDir)
 	if err != nil {
-		log.Debug().Msgf("Cannot read directory %s: %s", rpmImageDir, err)
+		log.Debug().Err(err).Msgf("Cannot read directory %s", rpmImageDir)
```
mbussolotto, line 115: Also here and below localization is missing. Maybe regex in `check_localizable` might be improved to find them

mbussolotto (approved):
just two small nitpick:
- localization is missing for some error message: I think regex in `check_localizable` are not catching them
- it's not completely clear for me when use L(), NL() or PL(). Maybe you can add a comment (or a paragraph in README.md) highlighting when to use one and when another (maybe with an example)

shared/podman/images.go
cbosdo: Again it's a debug: not localized

cbosdo:
> just two small nitpick:
> 
>     * localization is missing for some error message: I think regex in `check_localizable` are not catching them

I even added a check to not localized debug and traces there!

>     * it's not completely clear for me when use L(), NL() or PL(). Maybe you can add a comment (or a paragraph in README.md) highlighting when to use one and when another (maybe with an example)

Right! the idea is that `L()` is the simple one, `NL()` should be used where there is a plural form of the message and `PL()` is like `L()` with some context for translators. Remember that they don't read the code 😜

PR #294, opened 2024-05-06
configure the database to provide enough connections for our default setup

mcalmer:
## What does this PR change?

In the container we have already the default postgresql.conf file with the default configuration of 100 connections.
smdba autotuning works this was:
1. use the value provided as parameter on the command line
2. use the configured value from the current config (this match here and 100 is used)
3. use the internal default of 400

But internally smdba has a minimal connections of "200". As 100 is not high enough, smdba configure 200 in the container.

As the minimal required connections in our SUMA/Uyuni default setup is 386, we configure it now with 400.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/templates/pgsqlFinalizeScriptTemplate.go
+++ b/mgradm/shared/templates/pgsqlFinalizeScriptTemplate.go
@@ -14,7 +14,7 @@ set -e
 
 {{ if .RunAutotune }}
 echo "Running smdba system-check autotuning..."
-smdba system-check autotuning
+smdba system-check autotuning --max_connections=400
 {{ end }}
 echo "Starting Postgresql..."
 su -s /bin/bash - postgres -c "/usr/share/postgresql/postgresql-script start"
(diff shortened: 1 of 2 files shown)


srbarrios (approved):

mbussolotto:
> Question: is pgsqlFinalizeScriptTemplate.go only called on initial setup, or always on every container start?

`pgsqlFinalizeScriptTemplate.go` it's neither called during initial setup nor during container start: it's called only on `upgrade`.

cbosdo (approved):

mcalmer:
> > Question: is pgsqlFinalizeScriptTemplate.go only called on initial setup, or always on every container start?
> 
> `pgsqlFinalizeScriptTemplate.go` it's neither called during initial setup nor during container start: it's called only on `upgrade` (and `migrate`)

Hmm, than this change is wrong.
Who call `smdba` during setup?

mcalmer:
Ok, i need to change mgr-setup instead

mcalmer:
during migration we should use the configured values

mcalmer:
Replaced by https://github.com/uyuni-project/uyuni/pull/8700

PR #303, opened 2024-05-13
coco attestation container improvements

lucidd:
## What does this PR change?

This adds support for multiple instances of the confidential computing attestation container and a scale command to adjust the replicas.

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24238 https://github.com/SUSE/spacewalk/issues/24205

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge


mbussolotto (changes requested):
I left some comment, most of them are just about missing comments or return value.
Besides that:
- is it right to not have server attestation for uyuni on kubernetes?
- is there any better way to scale podman (maybe using `pod`?
cc @cbosdo

mgradm/cmd/scale/podman.go
```diff
@@ -0,0 +1,30 @@
...
+
+func podmanScale(
+	globalFlags *types.GlobalFlags,
+	flags *scaleFlags,
+	cmd *cobra.Command,
+	args []string,
+) error {
+	newReplicas := flags.Replicas
+	service := args[0]
+	if service == "uyuni-server-attestation" {
```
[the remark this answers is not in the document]
lucidd: i wasn't sure if we want to use the full service name here since its quite long or rather a shorter alias. I'm open for suggestions.

mgradm/cmd/status/podman.go
```diff
@@ -25,23 +26,18 @@ func podmanStatus(
 ) error {
 	// Show the status and that's it if the service is not running
 	if !podman.IsServiceRunning(podman.ServerService) {
-		if err := utils.RunCmdStdMapping(zerolog.DebugLevel, "systemctl", "status", "--no-pager", podman.ServerService); err != nil {
-			return fmt.Errorf(L("failed to get status of the server service: %s"), err)
+		utils.RunCmdStdMapping(zerolog.DebugLevel, "systemctl", "status", "--no-pager", podman.ServerService)
```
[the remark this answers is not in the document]
lucidd: not sure if checking the error in this case makes sense since systemctl returns non 0 status codes for cases that are not errors i.e 3 in case the service is not running. This currently leads to this message `Error: failed to get status of the server service: exit status 3` being printed at the end of a successful status run. If the underlying command has a real error there should still be the forwarded stderr printed for more context.

mgradm/cmd/status/podman.go
```diff
@@ -25,23 +26,18 @@ func podmanStatus(
...
-		return fmt.Errorf(L("failed to run spacewalk-service status: %s"), err)
-	}
-
-	if !podman.IsServiceRunning(podman.ServerAttestationService) {
-		if err := utils.RunCmdStdMapping(zerolog.DebugLevel, "systemctl", "status", podman.ServerAttestationService); err != nil {
-			return fmt.Errorf(L("failed to get status of the server service: %s"), err)
-		}
-		return nil
+	for i := 0; i < coco.CurrentReplicaCount(); i++ {
+		println() // add an empty line between the previous logs and this one
```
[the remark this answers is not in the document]
cbosdo: and it wouldn't make sense to log an empty message in the log file. BTW the logger logs to both the log file and stdout

mgradm/shared/coco/coco.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/uyuni-project/uyuni-tools/mgradm/shared/templates"
+	. "github.com/uyuni-project/uyuni-tools/shared/l10n"
+	"github.com/uyuni-project/uyuni-tools/shared/podman"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
+)
+
+// CurrentReplicaCount returns the current enabled replica count for a template service
+// name is the name of the service without the '.service' part.
+func CurrentReplicaCount(name string) int {
```
cbosdo, line 21: I would move this function to the `shared/podman/systemd.go` since that is generic enough to be used by other containers in the future

mgradm/shared/coco/coco.go
```diff
@@ -0,0 +1,92 @@
...
+	count := 0
+	for podman.ServiceIsEnabled(fmt.Sprintf("%s@%d", name, count)) {
+		count += 1
+	}
+	return count
+}
+
+// scales a templated systemd service to the requested number of replicas.
+// name is the name of the service without the '.service' part.
+func ScaleService(replicas int, name string) error {
```
cbosdo, line 31: Same for this one: this is generic enough to be moved to `systemd.go`.

mgradm/cmd/scale/kubernetes.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/uyuni-project/uyuni-tools/shared/types"
+)
+
+func kubernetesScale(
+	globalFlags *types.GlobalFlags,
+	flags *scaleFlags,
+	cmd *cobra.Command,
+	args []string,
+) error {
+	return errors.New(L("kubernetes not supported yet"))
```
cbosdo, line 23: We would need to think about the coco pod definition in server helm chart. There should be a deployment just for the coco attestation image and here all you would need to do is run `kubectl scale --replicas %d -n thenamespace deploy/coco-attestation` .

If you don't want to do it now, please add a card to add kubernetes support of the coco attestation image.

cbosdo:
>     * is it right to not have server attestation for uyuni on kubernetes?

It's not right, but we can afford doing it later, still I don't like skipping kubernetes parts for too long as that means we could overlook important differences with the podman implementation.

>     * is there any better way to scale podman (maybe using `pod`?

Sadly no, there is no better way. podman doesn't have an equivalent of kubernetes `deployment` or ` replicaSet`.

cbosdo (commented):
You'll see quite some changes after a rebase on master. The PR looks good, but I added suggestions for changes after the new l10n rework that has been merged.

mgradm/cmd/scale/scale.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.Flags().Int("replicas", 0, L("How many replicas of a container should be started."))
+}
+
+// NewCommand adjusts a containers replicas.
+func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
+	scaleCmd := &cobra.Command{
+		Use:     "scale",
+		GroupID: "management",
+		Short:   L("Adjust the replicas for container that supports it."),
+		Long:    L("Adjust the replicas for container that supports it."),
```
cbosdo, line 30: Shouldn't we add a list of containers that supports it in the Long help? I already see people complaining because `mgradm scale --replicas 2 uyuni-server`  doesn't work

mgradm/cmd/scale/podman.go
```diff
@@ -0,0 +1,30 @@
...
+	flags *scaleFlags,
+	cmd *cobra.Command,
+	args []string,
+) error {
+	newReplicas := flags.Replicas
+	service := args[0]
+	if service == podman.ServerAttestationService {
+		return podman.ScaleService(newReplicas, service)
+	}
+	return fmt.Errorf(L("Unknown service name: %s"), service)
```
cbosdo, line 29:
```suggestion
	return fmt.Errorf(L("service not allowing to be scaled: %s"), service)
```

mgradm/cmd/uninstall/podman.go
```diff
@@ -24,9 +27,16 @@ func uninstallForPodman(
 	// Force stop the pod
 	podman.DeleteContainer(podman.ServerContainerName, !flags.Force)
 
-	if podman.HasService(podman.ServerAttestationService) {
-		podman.UninstallService(podman.ServerAttestationService, !flags.Force)
-		podman.DeleteContainer(podman.ServerAttestationService, !flags.Force)
+	if err := podman.ScaleService(0, podman.ServerAttestationService); err != nil {
+		return fmt.Errorf(L("cannot delete confidential computing attestation instances: %s"), err)
```
cbosdo, line 31:
```suggestion
		return utils.Errorf(err, L("cannot delete confidential computing attestation instances"))
```

cbosdo (commented):
One forgotten `utils.Errorf()` and some suggestion on the supported containers error / help.

mgradm/cmd/scale/scale.go
lucidd: makes sense. how about 
`Adjust the replicas for containers that support it. Supported containers: uyuni-server-attestation` ?

cbosdo: You can use multiline strings using backticks. After thinking about the wording again, we should probably name those differently than containers: in the kubernetes world this would address a `deployement` or `replicaSet` and the `container` is a different concept. Service isn't much better since it is the thing that exposes the network ports of the pods in the cluster...

```suggestion
		Short:   L("Adjust the replicas for services that supports it."),
		Long:    L(`Adjust the replicas for services that supports it.

Supported services:
  - uyuni-server-attestation
`),
```

lucidd: it will always be confusing since the terminology of one backend will never match the other backend.

cbosdo: > it will always be confusing since the terminology of one backend will never match the other backend.

Yes, but you'll have multiple containers for the same coco attestation, so even for `podman` case that would be confusing as the real containers will be `coco-attestation-%d`

mgradm/cmd/scale/scale.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
+)
+
+type scaleFlags struct {
+	Backend  string
+	Replicas int
+}
+
+func addScaleFlags(cmd *cobra.Command) {
+	cmd.Flags().Int("replicas", 0, L("How many replicas of a container should be started."))
```
cbosdo, line 21:
```suggestion
	cmd.Flags().Int("replicas", 0, L("How many replicas of a service should be started."))
```

cbosdo (approved):

mbussolotto (approved):
LGTM

Issue #318, opened 2024-05-17
Command "mgradm support config" should save the support file with a date to avoid collisions


rjmateus:
The support config command should save the file and avoid collisions with other older support configs instead of overwrite. To archive this we can add the date to the export file.

File name should follow the same name structure as before: `SCC_<HOSTNAME>_DATE.txz`
example: `scc_stfpospap10_240412_1217.txz`

After extracting, the compress file, all files need to be located in a subdirectory with the same name as the file (without the extension). Currently it's extracting files that can lead to conflict if more the one support config is extracted in one folder.

admd:
@rjmateus I am moving this issue under GMC here https://github.com/SUSE/spacewalk/issues/24298 for now. It would be good if we can fix it for GMC. We still have a week for branch freeze.

admd:
PR has been merged. Feel free to close the issue.

PR #321, opened 2024-05-18
supportconfig file name with date to avoid conflicts

rjmateus:
## What does this PR change?

Implements: https://github.com/uyuni-project/uyuni-tools/issues/318

Change the behavior of -o flag to define the path and not the filename. 
Supportconfig filename contains the hostname and data, to avoid conflicts and identify the machine.

TODO: we need to change the way the tar file is made, and save all files in a directory, to avoid conflicts when exported by supports.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/318

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/support/config/extractor.go
+++ b/mgradm/cmd/support/config/extractor.go
@@ -11,6 +11,8 @@ import (
 	"os/exec"
 	"path"
 	"regexp"
+	"strings"
+	"time"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -86,7 +88,11 @@ func extract(globalFlags *types.GlobalFlags, flags *configFlags, cmd *cobra.Comm
 
 	// Pack it all into a tarball
 	log.Info().Msg(L("Preparing the tarball"))
-	tarball, err := utils.NewTarGz(flags.Output)
+
+	supportFileName := getSupportConfigFileSaveName()
+	supportFilePath := fmt.Sprintf("%s/%s", flags.Output, supportFileName)
+
+	tarball, err := utils.NewTarGz(supportFilePath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


mgradm/cmd/support/config/extractor.go
```diff
@@ -86,13 +88,17 @@ func extract(globalFlags *types.GlobalFlags, flags *configFlags, cmd *cobra.Comm
 
 	// Pack it all into a tarball
 	log.Info().Msg(L("Preparing the tarball"))
-	tarball, err := utils.NewTarGz(flags.Output)
+
+	supportFileName := getSupportConfigFileSaveName()
+	supportFilePath := fmt.Sprintf("%s/%s.tar.gz", flags.Output, supportFileName)
```
cbosdo, line 93:
```suggestion
	supportFilePath := path.Join(flags.Output, supportFileName)
```
Using `path.Join()` is better. Adding the `tar.gz` suffix to the result of `getSupportConfigFileSaveName()` would be more consistent.

mgradm/cmd/support/config/extractor.go
```diff
@@ -105,3 +111,14 @@ func getSupportConfigPath(out []byte) string {
...
+
+func getSupportConfigFileSaveName() string {
+	hostname_b, err := utils.RunCmdOutput(zerolog.DebugLevel, "hostname")
+	hostname := "localhost"
+	if err != nil {
+		log.Warn().Err(err).Msg(L("Unable to detect hostname, using localhost"))
+		hostname = strings.TrimSpace(string(hostname_b))
+	}
+	now := time.Now()
+	return fmt.Sprintf("scc_%s_%s", hostname, now.Format("20060102_1504"))
```
cbosdo, line 123:
```suggestion
	return fmt.Sprintf("scc_%s_%s.tar.gz", hostname, now.Format("20060102_1504"))
```

mgradm/cmd/support/config/extractor.go
```diff
@@ -86,13 +88,17 @@ func extract(globalFlags *types.GlobalFlags, flags *configFlags, cmd *cobra.Comm
...
+	supportFilePath := fmt.Sprintf("%s/%s.tar.gz", flags.Output, supportFileName)
+
+	tarball, err := utils.NewTarGz(supportFilePath)
 	if err != nil {
 		return err
 	}
 
 	for _, file := range files {
-		if err := tarball.AddFile(file, path.Base(file)); err != nil {
+		if err := tarball.AddFile(file, fmt.Sprintf("%s/%s", supportFileName, path.Base(file))); err != nil {
```
cbosdo, line 101:
```suggestion
		if err := tarball.AddFile(file, path.Join(supportFileName, path.Base(file))); err != nil {
```

cbosdo (changes requested):

mgradm/cmd/support/config/extractor.go
rjmateus: the variable `supportFileName` is also used to define the path for files inside the tar file. If we add the extension to that it will also land inside the tar file and the extract folder will have it on the name. This was the reason I added the extension outside

rjmateus: I will change the code to use path.join

cbosdo: > the variable `supportFileName` is also used to define the path for files inside the tar file. If we add the extension to that it will also land inside the tar file and the extract folder will have it on the name. This was the reason I added the extension outside

OK, I oversaw this, thanks for explaining

rjmateus: changed

mgradm/cmd/support/config/extractor.go
rjmateus: changed

cbosdo (approved):

Issue #336, opened 2024-05-24
Align hub XML-RPC enabling options with Coco attestation


cbosdo:
* Change `mgradm install podman` 's `--hubxmlrpc-enable`  into `--hubxmlrpc-replicas` with a default of `0` in order to align with the coco attestation image.
* Handle `hub-xmlrpc` service in `mgradm scale`

The maximum number of replicas for this service should be `1`

PR #338, opened 2024-05-28
add upgrade command support for coco service

lucidd:
## What does this PR change?

This adds upgrade command support for the coco service

## Test coverage
- No tests

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [x] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #340, opened 2024-05-28
Change --hubxmlrpc-enable to --hubxmlrpc-replicas

nadvornik:
## What does this PR change?

Change mgradm install podman  --hubxmlrpc-enable into --hubxmlrpc-replicas.
Add support for hub xmlrpc container to mgradm scale.
`mgradm install --hubxmlrpc-replicas 0` does not install the service at all - I think we don't want to install it on standalone and peripheral servers.

## Test coverage
- No tests: manual testing

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/336

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/restart/podman.go
+++ b/mgradm/cmd/restart/podman.go
@@ -5,6 +5,8 @@
 package restart
 
 import (
+	"errors"
+
 	"github.com/spf13/cobra"
 	"github.com/uyuni-project/uyuni-tools/shared/podman"
 	"github.com/uyuni-project/uyuni-tools/shared/types"
@@ -16,19 +18,8 @@ func podmanRestart(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	err := podman.RestartService(podman.ServerService)
-	if err != nil {
-		return err
-	}
-	if podman.HasService(podman.ServerAttestationService) {
-		if err := podman.RestartService(podman.ServerAttestationService); err != nil {
-			return err
-		}
-	}
-	if podman.HasService(podman.HubXmlrpcService) {
-		if err := podman.RestartService(podman.HubXmlrpcService); err != nil {
-			return err
-		}
-	}
-	return nil
+	err1 := podman.RestartService(podman.ServerService)
+	err2 := podman.RestartInstantiated(podman.ServerAttestationService)
+	err3 := podman.RestartInstantiated(podman.HubXmlrpcService)
+	return errors.Join(err1, err2, err3)
 }
--- a/mgradm/cmd/restart/podman.go
+++ b/mgradm/cmd/restart/podman.go
@@ -5,11 +5,10 @@
 package restart
 
 import (
-	"errors"
-
 	"github.com/spf13/cobra"
 	"github.com/uyuni-project/uyuni-tools/shared/podman"
 	"github.com/uyuni-project/uyuni-tools/shared/types"
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
 func podmanRestart(
@@ -21,5 +20,5 @@ func podmanRestart(
 	err1 := podman.RestartService(podman.ServerService)
 	err2 := podman.RestartInstantiated(podman.ServerAttestationService)
 	err3 := podman.RestartInstantiated(podman.HubXmlrpcService)
-	return errors.Join(err1, err2, err3)
+	return utils.JoinErrors(err1, err2, err3)
 }
--- a/mgradm/cmd/start/podman.go
+++ b/mgradm/cmd/start/podman.go
@@ -5,8 +5,9 @@
 package start
 
 import (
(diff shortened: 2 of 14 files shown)


mgradm/cmd/restart/podman.go
```diff
@@ -20,13 +22,13 @@ func podmanRestart(
 	if err != nil {
 		return err
 	}
-	if podman.HasService(podman.ServerAttestationService) {
-		if err := podman.RestartService(podman.ServerAttestationService); err != nil {
+	for i := 0; i < podman.CurrentReplicaCount(podman.ServerAttestationService); i++ {
+		if err := podman.RestartService(fmt.Sprintf("%s@%d", podman.ServerAttestationService, i)); err != nil {
```
rjmateus, line 26: I'm not sure about this one. for now we only have 0 or 1 replicas, so it would be safe. In the future, we will fail and stop the restart services for subsequent services if one fails, and I'm not sure what would be the side effects.

mgradm/cmd/start/podman.go
```diff
@@ -23,8 +23,8 @@ func podmanStart(
 			return err
 		}
 	}
-	if podman.HasService(podman.HubXmlrpcService) {
-		if err := podman.StartService(podman.HubXmlrpcService); err != nil {
+	for i := 0; i < podman.CurrentReplicaCount(podman.HubXmlrpcService); i++ {
```
rjmateus, line 26: I think this is also used by other services that doesn't have the scale part, like suse manager server service. Will it influence it?

mgradm/cmd/stop/podman.go
```diff
@@ -23,5 +23,10 @@ func podmanStop(
 			return err
 		}
 	}
+	for i := 0; i < podman.CurrentReplicaCount(podman.HubXmlrpcService); i++ {
```
rjmateus, line 26: same as podman start, other services are using this one, it may conflict.

rjmateus (commented):
there are some concerns at the start and stop methods for podman.

rjmateus:
please check the PR from CoCo, and see how Kevin have work around the start|stop for the replicas. I think it can follow a similar approach: https://github.com/uyuni-project/uyuni-tools/pull/338

mgradm/cmd/start/podman.go
nadvornik: I can't see any conflict.

nadvornik:
I have updated the PR with the coco changes https://github.com/uyuni-project/uyuni-tools/pull/338 and merged some code.

mgradm/cmd/start/podman.go
rjmateus: my bad, I miss look it.

shared/podman/systemd.go
```diff
@@ -184,6 +234,36 @@ func EnableService(service string) error {
 	return nil
 }
 
+// StartInstantiated starts all replicas.
+func StartInstantiated(service string) error {
+	for i := 0; i < CurrentReplicaCount(service); i++ {
+		if err := StartService(fmt.Sprintf("%s@%d", service, i)); err != nil {
+			return err
```
rjmateus, line 241: In this case is return an error when one of the replicas fail. Should we stack all error and return at the end, in case one of the replicas fails, it can continue to the other ones?
What your thought on this one?

nadvornik: This is copied from coco PR. Now it is used for both services. I am not sure if there was any reason why it was implemented this way for coco container.
For hub xmlrpc it does not matter, we allow max. 1 instance.

cbosdo: I agree with Ricardo: this is a point that has been overlooked in the coco implementation. That said there are not much chances to have the subsequent instances to start if one fails...

cbosdo (approved):

shared/podman/systemd.go
nadvornik: ok, fixed

shared/utils/utils.go
```diff
@@ -429,3 +429,18 @@ func Errorf(err error, message string, args ...any) error {
 	appended := fmt.Sprintf(message, args...) + ": " + err.Error()
 	return errors.New(appended)
 }
+
+// Join multiple errors.
+// Replacement for errors.Join which is not available in go 1.19.
```
cbosdo, line 434: why do we need a replacement for a missing bit in go 1.19 when the minimun go version is 1.20?

cbosdo (commented):
We probably don't need go 1.19 compatible code

shared/utils/utils.go
nadvornik: 1.19 is allowed in spec file on Red Hat distros. I don't want to break anything.

Issue #146, opened 2024-02-01
Add mgrpxy command to clear the squid cache


cbosdo:
In a lot of cases the squid proxy cache needs to be cleared. Providing an `mgrpxy` command for this would help.

PR #352, opened 2024-06-07
Add mgrpxy squid clear command

rjpmestre:
## What does this PR change?

This PR provides a new mgrpxy command so users can clear squid cache on demand.

## Test coverage
- No tests: manual tests

- [X] **DONE**

## Links

Issue(s): #146 

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgrpxy/cmd/status/kubernetes.go
+++ b/mgrpxy/cmd/status/kubernetes.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
+	"github.com/uyuni-project/uyuni-tools/shared"
 	"github.com/uyuni-project/uyuni-tools/shared/kubernetes"
 	. "github.com/uyuni-project/uyuni-tools/shared/l10n"
 	"github.com/uyuni-project/uyuni-tools/shared/types"
@@ -21,24 +22,14 @@ func kubernetesStatus(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	// Do we have an uyuni helm release?
-	clusterInfos, err := kubernetes.CheckCluster()
+	cnx := shared.NewConnection("kubectl", "", kubernetes.ProxyFilter)
+	namespace, err := cnx.GetNamespace("")
 	if err != nil {
-		return utils.Errorf(err, L("failed to discover the cluster type"))
-	}
-
-	kubeconfig := clusterInfos.GetKubeconfig()
-	if !kubernetes.HasHelmRelease("uyuni-proxy", kubeconfig) {
-		return errors.New(L("no uyuni-proxy helm release installed on the cluster"))
-	}
-
-	namespace, err := kubernetes.FindNamespace("uyuni-proxy", kubeconfig)
-	if err != nil {
-		return utils.Errorf(err, L("failed to find the uyuni-proxy deployment namespace"))
+		return err
 	}
 
 	// Is the pod running? Do we have all the replicas?
-	status, err := kubernetes.GetDeploymentStatus(namespace, "uyuni-proxy")
+	status, err := kubernetes.GetDeploymentStatus(namespace, kubernetes.ProxyApp)
 	if err != nil {
 		return utils.Errorf(err, L("failed to get deployment status"))
 	}
--- a/shared/connection.go
+++ b/shared/connection.go
@@ -6,6 +6,7 @@ package shared
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -29,18 +30,19 @@ type Connection struct {
 	backend          string
 	command          string
 	podName          string
-	podmanContainer  string
 	kubernetesFilter string
+	namespace        string
+	container        string
 }
 
 // Create a new connection object.
 // The backend is either the command to use to connect to the container or the empty string.
 //
 // The empty strings means automatic detection of the backend where the uyuni container is running.
-// podmanContainer is the name of a podman container to look for when detecting the command.
+// container is the name of a container to look for when detecting the command.
 // kubernetesFilter is a filter parameter to use to match a pod.
-func NewConnection(backend string, podmanContainer string, kubernetesFilter string) *Connection {
-	cnx := Connection{backend: backend, podmanContainer: podmanContainer, kubernetesFilter: kubernetesFilter}
+func NewConnection(backend string, container string, kubernetesFilter string) *Connection {
+	cnx := Connection{backend: backend, container: container, kubernetesFilter: kubernetesFilter}
 
 	return &cnx
 }
@@ -80,7 +82,7 @@ func (c *Connection) GetCommand() (string, error) {
 			for _, bin := range bins {
 				if _, err = exec.LookPath(bin); err == nil {
 					hasPodman = true
-					if checkErr := utils.RunCmd(bin, "inspect", c.podmanContainer, "--format", "{{.Name}}"); checkErr == nil {
+					if checkErr := utils.RunCmd(bin, "inspect", c.container, "--format", "{{.Name}}"); checkErr == nil {
 						c.command = bin
 						break
 					}
@@ -112,6 +114,59 @@ func (c *Connection) GetCommand() (string, error) {
 	return c.command, err
 }
 
+// GetNamespace finds the namespace of the running pod
+// appName is the name of the application to look for, if not provided it will be guessed based on the filter.
+// filters are additional filters to use to find the pod.
+func (c *Connection) GetNamespace(appName string, filters ...string) (string, error) {
+	// skip if namespace is already set
+	if c.namespace != "" {
+		return c.namespace, nil
+	}
+
+	command, cmdErr := c.GetCommand()
+	if cmdErr != nil {
+		return "", cmdErr
+	}
+
+	// skip if the command is not resolvable or does not target kubectl
+	if command != "kubectl" {
+		return c.namespace, nil
+	}
+
+	// if no appName is provided, we'll assume it based on its filter
+	if appName == "" {
+		switch c.kubernetesFilter {
+		case kubernetes.ProxyFilter:
+			appName = kubernetes.ProxyApp
+		case kubernetes.ServerFilter:
+			appName = kubernetes.ServerApp
+		}
+
+		if appName == "" {
+			return "", fmt.Errorf(L("coundn't find app name"))
+		}
+	}
+
+	// retrieving namespace from helm release
+	clusterInfos, clusterInfosErr := kubernetes.CheckCluster()
+	if clusterInfosErr != nil {
+		return "", utils.Errorf(clusterInfosErr, L("failed to discover the cluster type"))
+	}
+
+	kubeconfig := clusterInfos.GetKubeconfig()
+	if !kubernetes.HasHelmRelease(appName, kubeconfig) {
+		return "", fmt.Errorf(L("no %s helm release installed on the cluster"), appName)
+	}
+
+	var namespaceErr error
+	c.namespace, namespaceErr = extractNamespaceFromConfig(appName, kubeconfig)
+	if namespaceErr != nil {
+		return "", utils.Errorf(namespaceErr, L("failed to find the %s deployment namespace"), appName)
+	}
+
+	return c.namespace, nil
+}
+
 // GetPodName finds the name of the running pod.
 func (c *Connection) GetPodName() (string, error) {
 	var err error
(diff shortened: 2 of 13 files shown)


mbussolotto (changes requested):

shared/podman/utils.go
```diff
@@ -164,6 +164,24 @@ func DeleteVolume(name string, dryRun bool) error {
...
+func CreateVolume(name string, dryRun bool) error {
+	exists := isVolumePresent(name)
+	if !exists {
+		if dryRun {
+			log.Info().Msgf(L("Would run %s"), "podman volume create "+name)
+		} else {
+			log.Info().Msgf(L("Run %s"), "podman volume create "+name)
+			err := utils.RunCmd("podman", "volume", "create", name)
+			if err != nil {
+				log.Error().Err(err).Msgf(L("Failed to create volume %s"), name)
```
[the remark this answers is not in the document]
mbussolotto: you should do somethink like https://github.com/uyuni-project/uyuni-tools/blob/51722a4c1c7c03d38690e0d9db3578fc7b346f74/mgradm/cmd/install/podman/utils.go#L88

mgrpxy/cmd/squid/kubernetes.go
```diff
@@ -0,0 +1,36 @@
...
+	cmd *cobra.Command,
+	args []string,
+) error {
+	cnx := shared.NewConnection("kubectl", "", kubernetes.ProxyFilter)
+	podName, err := cnx.GetPodName()
+	if err != nil {
+		return utils.Errorf(err, L("failed to get pod name"))
+	}
+
+	err = utils.RunCmdStdMapping(zerolog.DebugLevel, "kubectl", "exec", podName, "-c", "squid", "--", "find", "/var/cache/squid", "-mindepth", "1", "-delete")
```
mbussolotto, line 30: I need an opinion from @cbosdo here :). We've never used `kubectl exec` until now, so I'm not sure if there's any limit or better way for doing it

cbosdo: We use it in a few places. I see two problems here:

* `Connection.Exec()` should be extended to work in the proxy too and use here.

* There is no namespace selected and thus the `default` will be picked while the proxy may be installed in another one. You'll need to detect the namespace first and then add it with `'-n', namespace` to the kubectl command options.

It also smells like `Connection.Exec()` and a few other places in our code don't handle the namespace correctly.

mgrpxy/cmd/squid/podman.go
```diff
@@ -0,0 +1,34 @@
...
+
+	if err := podman.StopService(podman.ProxyService); err != nil {
+		return err
+	}
+
+	if err := podman.DeleteVolume(volumeName, false); err != nil {
+		return err
+	}
+
+	if err := podman.CreateVolume(volumeName, false); err != nil {
```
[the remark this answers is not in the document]
cbosdo: Couldn't you handle the clear as done in Kubernetes? If that works it avoids volumes handling...

mgrpxy/cmd/squid/kubernetes.go
rjpmestre: I think i addressed your concerns. I ended up expanding the Connection struct so we can explicitly target a namepace and/or container.

shared/connection.go
```diff
@@ -42,6 +44,17 @@ func NewConnection(backend string, podmanContainer string, kubernetesFilter stri
 	return &cnx
 }
 
+// An explicit constructor for a kubernetes connection
+// Allows to specify:
+// - namespace: the namespace where the pod is running
+// - kubernetesFilter: a filter to find the pod
+// - kubernetesContainer: the container name.
+func NewKubernetesConnection(namespace string, kubernetesFilter string, kubernetesContainer string) *Connection {
```
cbosdo, line 52: I would probably have the `kubernetesContainer`  left as a parameter of the `Exec()` method so you could use the same connection to deal with multiple containers of the same pod. But that can be changed later if that hurts.

shared/connection.go
```diff
@@ -158,7 +171,22 @@ func (c *Connection) Exec(command string, args ...string) ([]byte, error) {
 
 	cmdArgs := []string{"exec", c.podName}
 	if cmd == "kubectl" {
-		cmdArgs = append(cmdArgs, "-c", "uyuni", "--")
+		namespace := c.namespace
+		if namespace == "" {
+			var namespaceErr error
+			namespace, namespaceErr = kubernetes.GetNamespace(c.kubernetesFilter)
```
cbosdo, line 177: You surely want to store the found namespace in the connection object...
```suggestion
			namespace, namespaceErr = kubernetes.GetNamespace(c.kubernetesFilter)
			c.namespace = namespace
```

shared/kubernetes/kubernetes.go
```diff
@@ -150,3 +151,28 @@ func GetSecret(secretName string, filter string) (string, error) {
 
 	return string(decoded), nil
 }
+
+// GetNamespace returns the namespace either for a uyuni or uyuni-proxy deployment.
+func GetNamespace(kubernetesFilter string) (string, error) {
```
cbosdo, line 156: You could even have that function in the `Connection` object itself.

mgrpxy/cmd/status/kubernetes.go
```diff
@@ -21,18 +21,7 @@ func kubernetesStatus(
...
-		return utils.Errorf(err, L("failed to discover the cluster type"))
-	}
-
-	kubeconfig := clusterInfos.GetKubeconfig()
-	if !kubernetes.HasHelmRelease("uyuni-proxy", kubeconfig) {
-		return errors.New(L("no uyuni-proxy helm release installed on the cluster"))
-	}
-
-	namespace, err := kubernetes.FindNamespace("uyuni-proxy", kubeconfig)
+	namespace, err := kubernetes.GetNamespace("uyuni-proxy")
```
cbosdo, line 24: I'm fine with such a change, but could you please kick away the other `FindNamespace()` uses to replace them with your new function?

rjpmestre: I was looking to find an elegant of doing but: 
- [HelmUnistall](https://github.com/uyuni-project/uyuni-tools/pull/352/files#diff-8666b3094e209ad406b9a5116fc2250b2e6666ffb4ef1ae6dfd2c2e9016085fcR65) uses it. 
- I did start by merging the content of both methods. It resulted in a method with a high cognitive complexity, hurting readability. 
For this reasons I just renamed the method for for something that goes with the method's specificity content, that is: to extract the namespace from a kubectl config file. 
Let me know what you think. I am under the impression you have quite a clear vision of what you'd like here and i'm just not being able to get it yet.

shared/connection.go
rjpmestre: Got you! Followed the same approach as [GetPodname](https://github.com/uyuni-project/uyuni-tools/pull/352/files#diff-f9195857a277d58e69d73592278875495daf2ddeacd42678cd2e64033b89fcc8L113-R158)

mgrpxy/cmd/status/kubernetes.go
cbosdo: It's strange that we have several ways of getting the used namespace...

shared/connection.go
cbosdo: > Got you! Followed the same approach as [GetPodName()](https://github.com/uyuni-project/uyuni-tools/pull/352/files#diff-f9195857a277d58e69d73592278875495daf2ddeacd42678cd2e64033b89fcc8L113-R158)

Yes, but `GetPodName()` stores the value in its field too :wink:

mgrpxy/cmd/status/kubernetes.go
rjpmestre: I gave it a go. Had to rework a bit the HelmUninstall method and it's callers. Let me know what you think of this approach

shared/connection.go
```diff
@@ -318,3 +397,32 @@ func chooseBackend[F interface{}](
 	// Should never happen if the commands are the same than those handled in GetCommand()
 	return nil, errors.New(L("no supported backend found"))
 }
+
+// extractNamespaceFromConfig extracts the namespace of a given application
+// from the Helm release information from an also given kubeconfig file (path).
```
cbosdo, line 402:
```suggestion
// from the Helm release information.
```
The kubeconfig path is just for the case it's installed in a non-standard place as k3s or rke2 are doing. If it's in `~/.kube/config` it can be empty.

shared/connection.go
```diff
@@ -109,6 +112,69 @@ func (c *Connection) GetCommand() (string, error) {
...
+
+	// try to get the namespace from the deployment
+	jsonpath := fmt.Sprintf("jsonpath={.items[?(@.metadata.name==\"%s\")].metadata.namespace}", appName)
+	args := []string{"get", "-A", "deploy", "-o", jsonpath}
+	args = append(args, filters...)
+
+	if namespace, _ := utils.RunCmdOutput(zerolog.DebugLevel, "kubectl", args...); len(namespace) != 0 {
+		c.namespace = string(namespace[:])
+		return c.namespace, nil
+	}
```
cbosdo, line 156: I think we can get rid of this attempt and directly try getting it from helm.

shared/connection.go
```diff
@@ -109,6 +112,59 @@ func (c *Connection) GetCommand() (string, error) {
...
+// filters are additional filters to use to find the pod.
+func (c *Connection) GetNamespace(appName string, filters ...string) (string, error) {
+	// skip if namespace is already set
+	if c.namespace != "" {
+		return c.namespace, nil
+	}
+
+	command, cmdErr := c.GetCommand()
+	if cmdErr != nil {
+		log.Fatal().Err(cmdErr)
```
cbosdo, line 126:
```suggestion
		return "", cmdErr
```

mgrpxy/cmd/squid/podman.go
```diff
@@ -0,0 +1,33 @@
...
+
+	if _, err := cnx.Exec("sh", "-c", "rm -rf /var/cache/squid/*"); err != nil {
+		return utils.Errorf(err, L("failed to remove cached data"))
+	}
+
+	if _, err := cnx.Exec("sh", "-c", "squid -z --foreground"); err != nil {
+		return utils.Errorf(err, L("failed to re-create the cache directories"))
+	}
+
+	return podman.EnableService(podman.ProxyService)
```
cbosdo, line 32: I'm not sure about this... In the running container you remove the cache files and recreate them... but enabling the service won't restart it... You may want `podman.RestartService(podman.ProxyService)` instead.

mgrpxy/cmd/squid/kubernetes.go
```diff
@@ -0,0 +1,33 @@
...
+
+func kubernetesSquidClear(
+	globalFlags *types.GlobalFlags,
+	flags *squidClearFlags,
+	cmd *cobra.Command,
+	args []string,
+) error {
+	cnx := shared.NewConnection("kubectl", "", kubernetes.ProxyFilter)
+
+	if _, err := cnx.ExecInKubernetesContainer("squid", "find", "/var/cache/squid", "-mindepth", "1", "-delete"); err != nil {
```
cbosdo, line 24: You have to read the function code to understand why you need it here and not the next exec. I would just expose a `Container` member in the Connection, set it before and run the regular `cnx.Exec()` everywhere. It would be more readable.

shared/connection.go
```diff
@@ -23,11 +24,13 @@ import (
...
-	podName          string
-	podmanContainer  string
-	kubernetesFilter string
+	backend             string
+	command             string
+	podName             string
+	podmanContainer     string
+	kubernetesFilter    string
+	namespace           string
+	kubernetesContainer string
```
cbosdo, line 33: Why not merge `podmanContainer` and `kubernetesContainer` into an exposed `Container` ?

mgrpxy/cmd/squid/kubernetes.go
rjpmestre: Mixing this and the idea of merging podmanContainer and kubernetesContainer, I think we wouldn't even need to expose it.

mgrpxy/cmd/squid/squid.go
```diff
@@ -0,0 +1,27 @@
...
+// NewCommand entry command for managing squid cache.
+// Setup for subcommand to clear (the cache).
+func NewCommand(globalFlags *types.GlobalFlags) *cobra.Command {
+	var squidCmd = &cobra.Command{
+		Use:   "squid",
+		Short: L("Manage Squid cache"),
+		Long:  L("Manage Squid cache"),
+		Run: func(cmd *cobra.Command, args []string) {
+			_ = cmd.Help()
+		},
```
cbosdo, line 22: I would not have `squid` in the command name, but something more generic like:

```suggestion
		Use:   "cache",
		Short: L("Manage proxy cache"),
		Long:  L("Manage proxy cache"),
		Run: func(cmd *cobra.Command, args []string) {
			_ = cmd.Help()
		},
```

Maybe later we'll add a `--squid` parameter to `mgrpxy cache clear` to target which cache to clear.  There are different caches in a proxy. Even though the squid one is usually to one to clear, we shouldn't close the door to the others.

rjpmestre: I see your point. Merged.

cbosdo (approved):

Issue #347, opened 2024-06-04
uyuniadm install with pre-existing volumes


rjmateus:
A valid scenario is to call the install and the volumes already exists. This is particular important in podman scenarios where we could uninstall without removing the volumes, or move the volumes to a different VM.

We must support calling the `mgradm install podman FQDN`.

With this call we must:
- We should detect that the volumes already exists, and print a warning message that the initial configuration parameter will be ignored
- Do not call the creat first user API method or if calling it detect that the first user was already set and do not finish with error.

We can call the set-up script, since it's protected at the volumes level with a check file.

PR #356, opened 2024-06-10
Allow installing with existing volumes

aaannz:
## What does this PR change?

This PR add check for existing uyuni-server service and if exists, refuse installation. It also add check for existing volume and display warning that not all installation options are preserved and make sure correct passwords are used.

At the end this PR adds new API call to try to verify if existing organization from pre existing volumes matches what would be created.

One small unrelated change is to use directly SUSEConnect instead of transactional-update --register. For plain status check, which is r/o operation, calling directly SUSEConnect is fine and we don't spend unnecessary time creating and discarding snapshots.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/347

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/install/podman/utils.go
+++ b/mgradm/cmd/install/podman/utils.go
@@ -6,6 +6,7 @@ package podman
 
 import (
 	"errors"
+	"fmt"
 	"os/exec"
 	"strings"
 
@@ -67,14 +68,18 @@ func installForPodman(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	flags.CheckParameters(cmd, "podman")
-	if _, err := exec.LookPath("podman"); err != nil {
-		return errors.New(L("install podman before running this command"))
-	}
-
 	inspectedHostValues, err := utils.InspectHost(false)
 	if err != nil {
 		return utils.Errorf(err, L("cannot inspect host values"))
+	}
+
+	if _, present := inspectedHostValues["host_uyuni_server"]; present {
+		return fmt.Errorf(L("Server is already initialized! Uninstall before attempting new installation or use upgrade command"))
+	}
+
+	flags.CheckParameters(cmd, "podman")
+	if _, err := exec.LookPath("podman"); err != nil {
+		return errors.New(L("install podman before running this command"))
 	}
 
 	fqdn, err := getFqdn(args)
--- a/mgradm/cmd/install/shared/shared.go
+++ b/mgradm/cmd/install/shared/shared.go
@@ -24,6 +24,13 @@ const setup_name = "setup.sh"
 
 // RunSetup execute the setup.
 func RunSetup(cnx *shared.Connection, flags *InstallFlags, fqdn string, env map[string]string) error {
+	// Containers should be running now, check storage if it is using volume from already configured server
+	preconfigured := false
+	if isServerConfigured(cnx) {
+		log.Warn().Msg(L("Server appears to be already configured. Installation will continue, but installation options may be ignored."))
+		preconfigured = true
+	}
+
 	tmpFolder := generateSetupScript(flags, fqdn, env)
 	defer os.RemoveAll(tmpFolder)
 
@@ -132,4 +139,8 @@ func boolToString(value bool) string {
 		return "Y"
 	}
 	return "N"
+}
+
+func isServerConfigured(cnx *shared.Connection) bool {
+	return cnx.TestExistenceInPod("/root/.MANAGER_SETUP_COMPLETE")
 }
--- a/mgradm/cmd/install/shared/shared.go
+++ b/mgradm/cmd/install/shared/shared.go
@@ -5,6 +5,8 @@
 package shared
 
 import (
+	"errors"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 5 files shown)


mgradm/cmd/install/podman/utils.go
```diff
@@ -67,16 +68,27 @@ func installForPodman(
...
 
+	/* Detect existing installation
+	- running server or configured, but not running server -> fail
+	- unconfigured, but volumes exists -> warn and continue
+	*/
+	if _, present := inspectedHostValues["host_uyuni_server"]; present {
+		return fmt.Errorf(L("Server is already initialized! Uninstall before attempting new installation or use upgrade command"))
+	}
+	if _, present := inspectedHostValues["host_uyuni_volume"]; present {
+		log.Warn().Msgf(L("Server volume is already present. Installation will continue, but installation options may not match. Make sure entered passwords match."))
```
cbosdo, line 84: Aren't the passwords simply ignored in such a case?
```suggestion
		log.Warn().Msgf(L("Server volume is already present. Installation will continue, but installation options may be ignored."))
```

aaannz: I am using the login and password to check if the first organization already exists and we are able to login.

cbosdo: > I am using the login and password to check if the first organization already exists and we are able to login.

Then may be this should be more explicit in the message. It starts to be quite long, we may have to cut it in two lines

```suggestion
		log.Warn().Msgf(L("Server volume is already present. Installation will continue, but installation options may be ignored. Make sure the administrator username and passwords match the previously set ones."))
```

rjmateus: Do you think is needed to check the password are the same? Isn't enough just a warning saying the provided data will be ignored?

mgradm/cmd/install/shared/shared.go
```diff
@@ -42,10 +43,28 @@ func RunSetup(cnx *shared.Connection, flags *InstallFlags, fqdn string, env map[
...
+			User:     flags.Admin.Login,
+			Password: flags.Admin.Password,
 		}
-		_, err := org.CreateFirst(&apiCnx, flags.Organization, &flags.Admin)
-		if err != nil {
-			return err
+
+		// Check if there is already admin user with given password and organization with same name
+		if _, err := api.Init(&apiCnx); err == nil {
+			if _, err := org.GetOrganizationDetails(&apiCnx, flags.Organization); err == nil {
```
rjmateus, line 52: What is the risk of just set-up and start the server if the credentials don't match? The use case proposed in here would be to stop the execution and provide the correct credentials and org to be able to set-up.
Is this needed? Can you foresee a scenario where the operator that are re-installing the server with existing volumes doesn't know the first user data?

mgradm/cmd/install/shared/shared.go
```diff
@@ -42,10 +43,28 @@ func RunSetup(cnx *shared.Connection, flags *InstallFlags, fqdn string, env map[
...
+				log.Debug().Err(err).Msg("Error returned by server")
+				return fmt.Errorf(L("Server organization already exists, but not with specified credentials"))
+			}
+		} else {
+			// We do not have any user existing, do not try to login
+			apiCnx = api.ConnectionDetails{
+				Server:   fqdn,
+				Insecure: true, // TODO Get the CA Cert and toggle this to false
+			}
+			_, err := org.CreateFirst(&apiCnx, flags.Organization, &flags.Admin)
```
rjmateus, line 64: Would make sense to always create the first user API method, and enhance it to return a different error code in case the first user already exists? that way we would be able to proper handle it and print the information.

aaannz: We return the info that it's already configured, however not in a code, but text which is dependent on locale I am afraid. HTTP code is 200 anyway.

mgradm/cmd/install/shared/shared.go
aaannz: The risk is only in scenario where something really does not work and we fail to detect that.

Scenarios:
1) happy path - new install and everything works
2) new install and something happens and deployment does not work
3) reusing volumes and everything works
4) reusing volumes and something does not work
  4a) user/pass mismatch, otherwise all is ok
  4b) image or data from volume are bad, server does not startup

So how to differentiate between 2, 4a and 4b scenarios?
We can maybe check if api calls returns HTTP200 -> if HTTP200, assume server is running, just auth problem so we'll report a warning, but return success?

mgradm/cmd/install/shared/shared.go
rjmateus: hum good point, I now got what I was overlooking. I was playing with the XML API and not the json over HTTP.
On XML RPC API we have the "faultCode", which identifies the error type that occurred. ON the Json over http looks like we are not mapping it.

aaannz:
Adapted:
- now we check for .MANAGER_SETUP_COMPLETE once container startup and issue warning if it is detected.
- check for existing organization now tries to distinguish between connection fail and some unknown failure - these will print error. If auth failure or mistyped org name is detected, only warning is issued, but we continue with success. And lastly if all matches, we just print info about it.

cbosdo (approved):
LGTM

rjmateus (approved):
The approach looks good to me. Thank you Ondrej

Issue #362, opened 2024-06-13
node-exporter port 9100 conflicts with traefik


nadvornik:
Uyuni container use port 9100 for node exporter. The same port  is used by traefik default configuration to exports it's own metrics, so there is a conflict and traefik fails to start:

```
2024/06/13 09:02:29 traefik.go:80: command traefik error: error while building entryPoint node-exporter: error preparing server: error opening listener: listen tcp :9100: bind: address already in use
```
The failure can be hidden because kubernetes keeps running the traefik with the old configuration. 

https://github.com/uyuni-project/uyuni-tools/pull/215
https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L704

@cbosdo @witekest

witekest:
Do we need _node exporter_ in Kubernetes deployments? It provides hardware metrics of the underlying node so I think it's out of scope for SUMA server monitoring.
It should be safe to remove _node exporter_ from the image for Kubernetes deployments, disable the service or remove the port mapping.
For container metrics we could switch to cAdvisor.

rjmateus:
For now, disable the port export for kubernetes. Next step is to try cAdvisor and decide what and how we should support. cAdvisor is part of packHUB only, we may need to support it on our side.

PR #364, opened 2024-06-17
Disable node exporter port for kubernetes, other fixes

nadvornik:
## What does this PR change?

- Disable node exporter port for kubernetes https://github.com/uyuni-project/uyuni-tools/issues/362
- Fix app name in start, stop and restart command on kubernetes
- Increase start timeout - 60s is not enough on my test machine

## Test coverage
- No tests: already covered

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/362

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/restart/kubernetes.go
+++ b/mgradm/cmd/restart/kubernetes.go
@@ -18,5 +18,5 @@ func kubernetesRestart(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	return kubernetes.Restart(kubernetes.ServerFilter)
+	return kubernetes.Restart(kubernetes.ServerApp)
 }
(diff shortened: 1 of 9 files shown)


mgradm/cmd/restart/kubernetes.go
```diff
@@ -18,5 +18,5 @@ func kubernetesRestart(
 	cmd *cobra.Command,
 	args []string,
 ) error {
-	return kubernetes.Restart(kubernetes.ServerFilter)
+	return kubernetes.Restart(kubernetes.ServerApp)
```
cbosdo, line 21: This is likely not to be the full fix: the Restart / Start / Stop functions are passing it to `getNode` and this one would need the full filter while `ReplicasTo` expects the filter value. Could you please align those two functions to complete the fix?

cbosdo (changes requested):

cbosdo (approved):

cbosdo:
@witekest I mentioned the use of cAdvisor in the comments regarding that port, but in the end we probably want to let the user decide: they could probably also use StackState or any other tool too...

PR #376, opened 2024-06-24
Fixes the way images URLs are computed

cbosdo:
## What does this PR change?

This fixes a few issues in the image URLs computing:

* The registry had to be copied from the global flags to the `ImageFlags`. This is quite error prone and already some images didn't have it.
* Don't use a different way to compute images in the proxy and the server tools.

This PR changes the `ComputeImage()` function to pass the registry and flags as parameters to avoid mistakes and move the tags computation there.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24849

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/utils/utils.go
+++ b/shared/utils/utils.go
@@ -182,16 +182,17 @@ func RemoveRegistryFromImage(imagePath string) string {
 }
 
 // ComputeImage assembles the container image from its name and tag.
-func ComputeImage(imageFlags types.ImageFlags, appendToName ...string) (string, error) {
-	if !strings.Contains(DefaultNamespace, imageFlags.Registry) {
-		log.Info().Msgf(L("Registry %[1]s would be used instead of namespace %[2]s"), imageFlags.Registry, DefaultNamespace)
-	}
+func ComputeImage(registry string, globalTag string, imageFlags types.ImageFlags, appendToName ...string) (string, error) {
 	name := imageFlags.Name
-	if !strings.Contains(imageFlags.Name, imageFlags.Registry) {
-		name = path.Join(imageFlags.Registry, RemoveRegistryFromImage(imageFlags.Name))
-		log.Info().Msgf(L("The image name provided is %[1]s and does not contains the registry %[2]s. The image name used will be %[3]s. You can set the flag --registry to change this behaviour."), imageFlags.Name, imageFlags.Registry, name)
+	if !strings.Contains(imageFlags.Name, registry) {
+		name = path.Join(registry, RemoveRegistryFromImage(imageFlags.Name))
 	}
-	tag := imageFlags.Tag
+
+	// Compute the tag
+	tag := globalTag
+	if imageFlags.Tag != "" {
+		tag = imageFlags.Tag
+	}
 
 	submatches := imageValid.FindStringSubmatch(name)
 	if submatches == nil {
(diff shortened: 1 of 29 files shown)


shared/utils/utils.go
```diff
@@ -166,16 +166,24 @@ func YesNo(question string) (bool, error) {
...
 	name := imageFlags.Name
-	if !strings.Contains(imageFlags.Name, imageFlags.Registry) {
-		name = path.Join(imageFlags.Registry, path.Base(imageFlags.Name))
-		log.Info().Msgf(L("The image name provided is %[1]s and does not contains the registry %[2]s. The image name used will be %[3]s. You can set the flag --registry to change this behaviour."), imageFlags.Name, imageFlags.Registry, name)
+	if !strings.Contains(imageFlags.Name, registry) {
+		name = path.Join(registry, path.Base(imageFlags.Name))
+		log.Info().Msgf(
+			L(`The image name provided is %[1]s and does not contains the registry %[2]s.
+			The image name used will be %[3]s, set the flag --registry to change this behaviour.`),
+			imageFlags.Name, registry, name)
```
aaannz, line 179: This is `Info` level message so displayed by default. I think that is quite chatty for that. Can't we just say `Using registry %s` without copying image path and help part?

shared/utils/utils.go
```diff
@@ -166,16 +166,24 @@ func YesNo(question string) (bool, error) {
 }
 
 // ComputeImage assembles the container image from its name and tag.
-func ComputeImage(imageFlags types.ImageFlags, appendToName ...string) (string, error) {
-	if !strings.Contains(DefaultNamespace, imageFlags.Registry) {
-		log.Info().Msgf(L("Registry %[1]s would be used instead of namespace %[2]s"), imageFlags.Registry, DefaultNamespace)
+func ComputeImage(registry string, globalTag string, imageFlags types.ImageFlags, appendToName ...string) (string, error) {
```
aaannz, line 169: It's not clear to me why having extra `registry` string when registry is already part of `imageFlags`?

aaannz: Or rather, why removing registry from imageFlags?

shared/utils/utils.go
cbosdo: I'm not convinced about the usefulness of this log either... I just reformatted it.

rjmateus:
We need to check how this PR will interfere with the logic merged on this PR: https://github.com/uyuni-project/uyuni-tools/pull/381

We should keep the logic that was implemented in that one, and defined in: https://github.com/uyuni-project/uyuni-tools/issues/378

cbosdo:
> We need to check how this PR will interfere with the logic merged on this PR: #381
> 
> We should keep the logic that was implemented in that one, and defined in: #378

I completely dropped the commit splitting the `--registry` parameter in two: I'm washing my hands off that decision...

aaannz (approved):
LGTM.

For clarification, this PR also contains this one, right? https://github.com/uyuni-project/uyuni-tools/pull/392/

cbosdo:
> LGTM.
> 
> For clarification, this PR also contains this one, right? #392

Yes it does. I can try to remove that commit though

Issue #181, opened 2024-02-27
Develop a prepare migration command


rjmateus:
For large deployments, especially with a huge number of packages and big database, migration can be complex and take a long time.

We should consider if we should have a prepare migration command which would rsync some folder that we know that are big, like:
- packages
- cache
- database

Since it's using rsync, it can be called several time without any problem, and with the source server running.

When the migration happens, we would stop the services on the source server, and run the migration command, which would rsync everything again and make the needed transformations.

cbosdo:
I would suggest to add `--prepare` flag to the `mgradm migrate [podman|kubernetes]` commands that only runs the `rsync` parts of the migration script.

admd:
As a user, when I run this command, I expect it to execute in the background with clear informational messages indicating that it will not perform the actual migration but will move some easily transferable data to save time during the actual migration. It would be ideal if users could check the status of this process though.

cbosdo:
> As a user, when I run this command, I expect it to execute in the background with clear informational messages indicating that it will not perform the actual migration but will move some easily transferable data to save time during the actual migration. It would be ideal if users could check the status of this process though.

Why in the background? The user can always run this in a `screen` if needed.

admd:
Ok yes, it should be fine to not run in background.

But then my question becomes, do we show what's happening behind the scene when running this command? We need to show some information or progress or anything. We cannot just run a command and then it just stuck there with no feedback to the user if something is happening or not. If we cover that part, we are good.

nadvornik:
The user can see the output of rsync command - a list of synced files.

cbosdo:
> The user can see the output of rsync command - a list of synced files.

Shouldn't we add `-P` to `rsync` to show the progress? or is what we have now good enough?

PR #377, opened 2024-06-24
Add --prepare option to migrate command

nadvornik:
## What does this PR change?

This PR adds `--prepare` option to `migrate` command.
With this option, the migration does not stop the services on source server and ends after syncing
the data.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/181

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/migrate/kubernetes/utils.go
+++ b/mgradm/cmd/migrate/kubernetes/utils.go
@@ -52,7 +52,7 @@ func migrateToKubernetes(
 	sshConfigPath, sshKnownhostsPath := migration_shared.GetSshPaths()
 
 	// Prepare the migration script and folder
-	scriptDir, err := adm_utils.GenerateMigrationScript(fqdn, flags.User, true)
+	scriptDir, err := adm_utils.GenerateMigrationScript(fqdn, flags.User, true, flags.Prepare)
 	if err != nil {
 		return utils.Errorf(err, L("failed to generate migration script"))
 	}
@@ -88,6 +88,11 @@ func migrateToKubernetes(
 	// Run the actual migration
 	if err := adm_utils.RunMigration(cnx, scriptDir, "migrate.sh"); err != nil {
 		return utils.Errorf(err, L("cannot run migration"))
+	}
+
+	if flags.Prepare {
+		log.Info().Msg(L("Migration prepared. Run the 'migrate' command without '--prepare' to finish the migration."))
+		return nil
 	}
 
 	tz, oldPgVersion, newPgVersion, err := adm_utils.ReadContainerData(scriptDir)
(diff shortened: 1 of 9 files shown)


mgradm/cmd/migrate/kubernetes/utils.go
```diff
@@ -90,6 +90,11 @@ func migrateToKubernetes(
 		return utils.Errorf(err, L("cannot run migration"))
 	}
 
+	if flags.Prepare {
+		log.Info().Msg(L("Migration prepared. Run the 'migrate' command without '--prepare' to finish the migration."))
```
cbosdo, line 94: Should we scale to 0 after this? Is the migration without `--prepare` coping well with the existing resources?

mgradm/cmd/migrate/podman/utils.go
```diff
@@ -35,10 +35,14 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *podmanMigrateFlags,
 	sshAuthSocket := migration_shared.GetSshAuthSocket()
 	sshConfigPath, sshKnownhostsPath := migration_shared.GetSshPaths()
 
-	tz, oldPgVersion, newPgVersion, err := podman.RunMigration(serverImage, flags.Image.PullPolicy, sshAuthSocket, sshConfigPath, sshKnownhostsPath, sourceFqdn, flags.User)
+	tz, oldPgVersion, newPgVersion, err := podman.RunMigration(serverImage, flags.Image.PullPolicy, sshAuthSocket, sshConfigPath, sshKnownhostsPath, sourceFqdn, flags.User, flags.Prepare)
 	if err != nil {
 		return utils.Errorf(err, L("cannot run migration script"))
 	}
+	if flags.Prepare {
+		log.Info().Msg(L("Migration prepared. Run the 'migrate' command without '--prepare' to finish the migration."))
```
cbosdo, line 43: Same here: should we stop the service here? Is `migrate` behaving correctly with the existing resources?

cbosdo (commented):
LGTM, with some questions about whether to stop the services?

mgradm/cmd/migrate/podman/utils.go
nadvornik: There is nothing to stop for podman. The container exists only while the migration script is running.

mgradm/cmd/migrate/kubernetes/utils.go
nadvornik: I have to test the kubernetes part.

rjmateus:
I if we set the prepare command we should stop just after the rsync of each volume. All the config changes (and probably the distribution copy) should be done only on final migrate. 
The goal is to start copy the major data folder to speed migration.

nadvornik:
I think it is safe to let the migration scrip run till the end. All the config changes it does will be overwritten by the final run, but it can help to spot potential problems earlier, before stopping the old server.

nadvornik:
I have added kubernetes fixes -  scale to 0 and skip traefik installation.

cbosdo (approved):

PR #379, opened 2024-06-25
Do not rely on hardcoded registries

aaannz:
## What does this PR change?

This commit instead of removing well-known registries, removes any fqdn when computing image name

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): none

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/podman/images.go
+++ b/shared/podman/images.go
@@ -9,8 +9,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/rs/zerolog"
@@ -21,11 +23,6 @@ import (
 )
 
 const rpmImageDir = "/usr/share/suse-docker-images/native/"
-
-var registries = []string{
-	"registry.suse.com",
-	"registry.opensuse.com",
-}
 
 // Ensure the container image is pulled or pull it if the pull policy allows it.
 //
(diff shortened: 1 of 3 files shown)


shared/podman/images.go
```diff
@@ -73,20 +70,24 @@ func PrepareImage(image string, pullPolicy string, args ...string) (string, erro
...
-			tag = "latest"
-			if len(parts) > 1 {
-				tag = parts[1]
-			}
-			rpmImageFile = parts[0]
-			return rpmImageFile, tag
-		}
+	pattern := regexp.MustCompile(`^https?://|^docker://|^oci://`)
+	if pattern.FindStringIndex(image) == nil {
+		image = "docker://" + image
```
rjmateus, line 75: brilliant trick :+1:

rjmateus (approved):
LGTM
Thank you for taking care of it

shared/podman/images.go
```diff
@@ -73,20 +70,24 @@ func PrepareImage(image string, pullPolicy string, args ...string) (string, erro
...
 	}
-	return "", ""
+	url, err := url.Parse(image)
```
cbosdo, line 77: Does this behave correctly if the server is addressed using a hostname rather than an FQDN?

aaannz: Yes. It assumes everything before first `/` as authority where authority is potential `user@host` string from which it takes host.

However I think that after release we can refactor this to use Vladimirs more direct approach from https://github.com/uyuni-project/uyuni-tools/pull/381 instead of pulling `net/url` dependency.

Issue #378, opened 2024-06-24
Define --registry flag behaviour


rjmateus:
For public cloud and internal registry usage, we must have a way to overwrite registry to where we want to pull the images from.
This card aims to agree on how to make it, and document what should be implemented.

For podman we always want to use the local RPM image if available. Local images are located at `/usr/share/suse-docker-images/native`. 
The metadata file will have the image name and available tags. The name in the RPM image is independent from the registry (example: `suse-manager-5.0-x86_64-server`), and is computed from the image namespace defined on SCC.

example:
   Image to use: `registry.suse.com/suse/manager/5.0/x86_64/server`
   Will look at the file system by image name `suse-manager-5.0-x86_64-server`

Our goals are:
- Allow to set a different registry for cloud and private registry.
- Always look for the local image (RPM) before trying to download it as much as possible

# Proposed Solution

Each image used in the deployment will have its configuration as we have now. Example: `--image`, `--coco-image`, `--hubxmlrpc-image`.
Those properties will be setting the image path/namespace, and can also set the tag name.

The registry flag will set the prefix. It can be just the registry FQDN, or also include some path prefix. Example:
- `--registry=myreg.com`
- `--registry=myreg.com/prefix/path/`

We always give priority to local images, but searching with the image path defined on image specific path flags. The values on those flags will be different on SUMA or Uyuni. The registry flag will affect only the online image pull.

If the user defines only the registry all images are affected with the same prefix.
If they want to change the default namespace, users should define it for each image one by one, with the correct location. Reason: if they start to play around with names, we cannot know how imaginative they can get, so define them individually.

Scenarios:
- Cloud registry: Only set the registry, should use the same image namespace and all should work with local lookup
- Private registry:
   - Keep same namespace: same as cloud, set registry flag and all should work with local lookup
   - Change namespace or tags: Define the registry (with or without prefix) and then for each image define the path.

## Examples

Default for `--registry`: registry.suse.com
Default image path `--image`: suse/manager/5.0/x86_64/server
Default image tag `--tag`: 5-0

### No flags defined

-- local look-up with
    - ImageName: suse/manager/5.0/x86_64/server
    - Tag: 5.0.0
- If fail, fallback to remote look-up with
    - Image: registry.suse.com/suse/manager/5.0/x86_64/server
    - Tag: 5.0.0

### Set the Registry

registry=cloud.com

Look up locally with the default image path. Remote will use the defined registry + the image path

-- local look-up with
[...]

PR #381, opened 2024-06-26
Define --registry flag behaviour (bsc#1226793)

nadvornik:
## What does this PR change?

This is a minimal change to implement the --registry flag behaviour as described in https://github.com/uyuni-project/uyuni-tools/issues/378 , except for default values.

The defaults are not changed - `--image` contains the registry and `--registry` is empty.
This PR has an effect only if non-empty `--registry` is specified.

Changing the defaults would require much more changes and testing.

## Test coverage
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s):
https://github.com/uyuni-project/uyuni-tools/issues/378
https://github.com/SUSE/spacewalk/issues/24638
https://bugzilla.suse.com/show_bug.cgi?id=1226793

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/utils/utils.go
+++ b/shared/utils/utils.go
@@ -165,6 +165,22 @@ func YesNo(question string) (bool, error) {
 	}
 }
 
+// Remove registry fqdn from image path.
+func RemoveRegistryFromImage(imagePath string) string {
+	separator := "://"
+	index := strings.Index(imagePath, separator)
+	if index != -1 {
+		imagePath = imagePath[index+len(separator):]
+	}
+
+	parts := strings.Split(imagePath, "/")
+	if strings.Contains(parts[0], ".") || strings.Contains(parts[0], ":") || index != -1 {
+		// first part is a registry fqdn
+		parts = parts[1:]
+	}
+	return strings.Join(parts, "/")
+}
+
 // ComputeImage assembles the container image from its name and tag.
 func ComputeImage(imageFlags types.ImageFlags, appendToName ...string) (string, error) {
 	if !strings.Contains(DefaultNamespace, imageFlags.Registry) {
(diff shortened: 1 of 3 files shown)


shared/utils/utils.go
```diff
@@ -165,14 +165,30 @@ func YesNo(question string) (bool, error) {
 	}
 }
 
+// Remove registry fqdn from image path.
+func RemoveRegistryFromImage(imagePath string) string {
```
rjmateus, line 169: Ondrej have implemented a similar approach for this in here: https://github.com/uyuni-project/uyuni-tools/blob/main/shared/podman/images.go#L73-L82

That implementation looks much more straightforward to me. I think we can extract that piece of code to a function and reuse it in here.

aaannz (approved):
Nice!

Maybe you can adapt `shared/podman/images.go` to use your `RemoveRegistryFromImage` instead of that net/url magic.

Also maybe add one unit test testcase having registry with `http://` scheme and also image with it as well.

rjmateus (changes requested):
Thank you for getting this solution, it looks great to me. The change on the default I think is safe to make it latter, since with this improvement is not urgent anymore.

shared/utils/utils.go
rjmateus: If you prefer your implementation, I think it's algo great. But I would prefer to have just one implementation of the registry removal

rjmateus: Now that I look closer to Ondrej implementation, I think we cannot use that code in here, since it assumes the registry fqdn is already part of the image name, which is not the case at this stage.

rjmateus (approved):

PR #382, opened 2024-07-01
Prefer local images in all migration steps (bsc#1227244)

rjmateus:
## What does this PR change?

Fix: https://bugzilla.suse.com/show_bug.cgi?id=1227244

The migration process has several different steps that need to use images. The image path, and if it should be local or remote was done in the first to steps but not on the next ones. Move the server image computation to outside, and passed that as an argument.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): #

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/migrate/podman/utils.go
+++ b/mgradm/cmd/migrate/podman/utils.go
@@ -31,11 +31,29 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *podmanMigrateFlags, 
 		return utils.Errorf(err, L("cannot compute image"))
 	}
 
+	// FIXME all this code should be centralized. Now it being called in several different places.
+	inspectedHostValues, err := utils.InspectHost(false)
+	if err != nil {
+		return utils.Errorf(err, L("cannot inspect host values"))
+	}
+
+	pullArgs := []string{}
+	_, scc_user_exist := inspectedHostValues["host_scc_username"]
+	_, scc_user_password := inspectedHostValues["host_scc_password"]
+	if scc_user_exist && scc_user_password {
+		pullArgs = append(pullArgs, "--creds", inspectedHostValues["host_scc_username"]+":"+inspectedHostValues["host_scc_password"])
+	}
+
+	preparedImage, err := podman_utils.PrepareImage(serverImage, flags.Image.PullPolicy, pullArgs...)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


mgradm/shared/podman/podman.go
```diff
@@ -199,25 +199,8 @@ func RunMigration(serverImage string, pullPolicy string, sshAuthSocket string, s
...
-	}
-
-	preparedImage, err := podman.PrepareImage(serverImage, pullPolicy, pullArgs...)
-	if err != nil {
-		return "", "", "", err
-	}
-
 	log.Info().Msg(L("Migrating server"))
-	if err := podman.RunContainer("uyuni-migration", preparedImage, utils.ServerVolumeMounts, extraArgs,
+	if err := podman.RunContainer("uyuni-migration", serverImage, utils.ServerVolumeMounts, extraArgs,
```
aaannz, line 203: pullPolicy is not used now, is it? So rather then change `preparedImage` -> `serverImage`, how about changing method arguments to `preparedImage` and removign `pullPolicy` option?

mgradm/cmd/migrate/podman/utils.go
```diff
@@ -31,11 +31,29 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *podmanMigrateFlags,
 		return utils.Errorf(err, L("cannot compute image"))
 	}
 
+	// FIXME all this code should be centralized. Now it being called in several different places.
```
aaannz, line 34: We have this centralized in one of our million of shared dirs - https://github.com/uyuni-project/uyuni-tools/blob/main/shared/podman/utils.go#L176

So should be extracted there.

And no, I don't know why we have 2 Inspect methods :/

aaannz (approved):
LGTM

One nitpick about args naming.

mgradm/cmd/migrate/podman/utils.go
rjmateus: sure, but in the expect, it doesn't return the `preparedImage`, and that is one of the thing we need, and what should be used to deploy. 
We where calling the prepare image name in several places, and missing in others

mgradm/shared/podman/podman.go
rjmateus: Good catch, I have change it according

PR #386, opened 2024-07-04
Fix random error during installation of CA certificate (bsc#1227245)

nadvornik:
## What does this PR change?

When a container is started via systemd, there is a short period when
GetPodName() can fail or GetPodName() succeeds and the following command fails.

It was reported on migration, but I think it can happen on install too. It is more likely on migration, probably because of large writes.

This PR fixes it.

## Test coverage
- No tests: already covered

- [ ] **DONE**

## Links

Issue(s): 
https://github.com/SUSE/spacewalk/issues/24705
https://bugzilla.suse.com/show_bug.cgi?id=1227245

- [ ] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/connection.go
+++ b/shared/connection.go
@@ -169,13 +169,43 @@ func (c *Connection) Exec(command string, args ...string) ([]byte, error) {
 	return utils.RunCmdOutput(zerolog.DebugLevel, cmd, cmdArgs...)
 }
 
+// WaitForContainer waits up to 10 sec for the container to appear.
+func (c *Connection) WaitForContainer() error {
+	for i := 0; i < 10; i++ {
+		podName, err := c.GetPodName()
+		if err != nil {
+			log.Warn().Err(err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		args := []string{"exec", podName}
+		command, err := c.GetCommand()
+		if err != nil {
+			log.Fatal().Err(err)
+		}
+
+		if command == "kubectl" {
+			args = append(args, "--")
+		}
+		args = append(args, "true")
+		err = utils.RunCmd(command, args...)
+		if err == nil {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return errors.New(L("container didn't start within 10s."))
+}
+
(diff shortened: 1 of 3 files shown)


rjmateus (approved):
lgtm. thank you for the solution

shared/connection.go
```diff
@@ -169,13 +169,43 @@ func (c *Connection) Exec(command string, args ...string) ([]byte, error) {
 	return utils.RunCmdOutput(zerolog.DebugLevel, cmd, cmdArgs...)
 }
 
+// WaitForContainer waits up to 10 sec for the container to appear.
+func (c *Connection) WaitForContainer() error {
+	for i := 0; i < 10; i++ {
+		podName, err := c.GetPodName()
+		if err != nil {
+			log.Warn().Err(err)
```
aaannz, line 177: I would mark this as Debug. This would make our users only waste time asking what is happening.

shared/connection.go
```diff
@@ -169,13 +169,43 @@ func (c *Connection) Exec(command string, args ...string) ([]byte, error) {
...
+}
+
 // WaitForServer waits at most 60s for multi-user systemd target to be reached.
 func (c *Connection) WaitForServer() error {
 	// Wait for the system to be up
 	for i := 0; i < 60; i++ {
 		podName, err := c.GetPodName()
 		if err != nil {
-			log.Fatal().Err(err)
+			log.Warn().Err(err)
```
aaannz, line 206: Same here.

aaannz (approved):
LGTM, I only think we can do this wait little less scary.

shared/connection.go
```diff
@@ -169,13 +169,43 @@ func (c *Connection) Exec(command string, args ...string) ([]byte, error) {
...
+		podName, err := c.GetPodName()
+		if err != nil {
+			log.Warn().Err(err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		args := []string{"exec", podName}
+		command, err := c.GetCommand()
+		if err != nil {
+			log.Fatal().Err(err)
```
cbosdo, line 184: better return the error rather than `log.Fatal()`

cbosdo (approved):

PR #391, opened 2024-07-08
Use the basename of the SQL file as filename in the container

cbosdo:
## What does this PR change?

To make sure there is no slash in the path of the SQL file in the container use the basename to compute the temporary filename.

## Test coverage
- No tests: executed commands cannot be unit tested yet

- [X] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24745

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/support/sql/sql.go
+++ b/mgradm/cmd/support/sql/sql.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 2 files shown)


mcalmer (approved):

mgradm/cmd/support/sql/sql.go
```diff
@@ -27,7 +28,7 @@ import (
 func prepareSource(args []string, cnx *shared.Connection) (string, error) {
 	source := "-"
 	if len(args) > 0 {
-		source = args[0]
+		source = path.Base(args[0])
 		if !utils.FileExists(source) {
 			return "", fmt.Errorf(L("source %s does not exists"), source)
 		}
```
aaannz, line 34: Are you sure this is correct here? I mean you can call `mgradm support sql /tmp/file.sql`

We need remove them only after this check.

cbosdo: indeed... my test was partial

cbosdo: now the variables are more cleanly separated: it'll be much better

PR #394, opened 2024-07-09
enforce cobbler setup (bsc#1226847)

mbussolotto:
## What does this PR change?

- validate FQDN
- check if FQDN is resolvable
- collect return code of `mgr-setup`
it requires this change on `uyuni` https://github.com/uyuni-project/uyuni/pull/9041

`IsWellFormedFQDN` and `IsValidFQDN` should prevents errors, but in case this function passed, this is how setup script fails if cobble is not setup correctly:
```
6:28PM INF * Setting up Cobbler..
6:28PM INF Could not setup cobbler  Exit value: 1.
Please examine /var/log/rhn/rhn_installation.log for more information.
6:28PM INF ERROR: spacewalk-setup failed
Error: error running the setup script: exit status 1
```

## Test coverage

- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24650
Related to : https://github.com/uyuni-project/uyuni/pull/9041

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/templates/mgrSetupScriptTemplate.go
+++ b/mgradm/shared/templates/mgrSetupScriptTemplate.go
@@ -20,10 +20,12 @@ echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:800
 echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:8002,server=y,suspend=n" ' >> /usr/share/rhn/config-defaults/rhn_search_daemon.conf
 {{- end }}
 
-/usr/lib/susemanager/bin/mgr-setup -s -n
+RESULT=$(/usr/lib/susemanager/bin/mgr-setup -s -n)
 
 # clean before leaving
-rm $0`
+rm $0
+exit $RESULT
+`
 
 // MgrSetupScriptTemplateData represents information used to create setup script.
 type MgrSetupScriptTemplateData struct {
--- a/mgradm/shared/templates/mgrSetupScriptTemplate.go
+++ b/mgradm/shared/templates/mgrSetupScriptTemplate.go
@@ -20,7 +20,8 @@ echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:800
 echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:8002,server=y,suspend=n" ' >> /usr/share/rhn/config-defaults/rhn_search_daemon.conf
 {{- end }}
 
-RESULT=$(/usr/lib/susemanager/bin/mgr-setup -s -n)
+/usr/lib/susemanager/bin/mgr-setup -s -n
+RESULT=$?
 
 # clean before leaving
 rm $0
--- a/shared/utils/utils.go
+++ b/shared/utils/utils.go
@@ -36,6 +36,8 @@ const prompt_end = ": "
 
 var prodVersionArchRegex = regexp.MustCompile(`suse\/manager\/.*:`)
 var imageValid = regexp.MustCompile("^((?:[^:/]+(?::[0-9]+)?/)?[^:]+)(?::([^:]+))?$")
+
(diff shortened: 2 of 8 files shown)


shared/utils/utils.go
```diff
@@ -22,6 +22,7 @@ import (
 	"syscall"
 	"unicode"
 
+	"github.com/go-playground/validator"
```
aaannz, line 25: Are we going to use this deps for something else?

For me this seems like quite a big overkill for one regexp validation.

Validator is MIT license, so we can copy that regexp with attribution from https://github.com/go-playground/validator/blob/master/regexes.go#L58

aaannz (commented):
Personally I am against pulling new dependency just for this, but seems like you have more plans with it 😄

The original bug was also about that our install script exits successfully despite wrong setup. Is that something we can catch in our tools or require change on uyuni side?

shared/utils/utils.go
mbussolotto: make sense. Let me try to avoid to sue the library and add just the regex

shared/utils/utils.go
```diff
@@ -35,6 +36,7 @@ const prompt_end = ": "
 
 var prodVersionArchRegex = regexp.MustCompile(`suse\/manager\/.*:`)
 var imageValid = regexp.MustCompile("^((?:[^:/]+(?::[0-9]+)?/)?[^:]+)(?::([^:]+))?$")
+var fqdnValid = regexp.MustCompile(`^([a-zA-Z0-9]{1}[a-zA-Z0-9-]{0,62})(\.[a-zA-Z0-9]{1}[a-zA-Z0-9-]{0,62})*?(\.[a-zA-Z]{1}[a-zA-Z0-9]{0,62})\.?$`)
```
aaannz, line 39: Personally I would add a attribution link.
e.g. `// Taken from https://github.com/go-playground/validator/blob/2e1df48/regexes.go#L58`

aaannz (approved):
LGTM with one nitpick.

mgradm/cmd/install/shared/shared.go
```diff
@@ -32,6 +32,8 @@ func RunSetup(cnx *shared.Connection, flags *InstallFlags, fqdn string, env map[
 	}
 
 	err := adm_utils.ExecCommand(zerolog.InfoLevel, cnx, "/tmp/setup.sh")
+	_ = adm_utils.ExecCommand(zerolog.InfoLevel, cnx, "rm /tmp/setup.sh")
```
cbosdo, line 35: Why removing the error code here? Is this a dev leftover? and the error test after this would not behave as expected.

cbosdo (changes requested):

mgradm/cmd/install/shared/shared.go
mbussolotto: mm actually this was intentional:
- I removed `rm $0` from the setup script: in this way the return code of the setup script is the return code of `mgr-setup`
- I added this line to remove the script and we want to remove it because this action should be perform this action also if the setup fails, that's why it is performed before checking the error of the setup script
- I ignored the error in `rm /tmp/setup.sh` because I don't think the installation should fails if we have problems just on removing the script.

At least this was the intention :D do you have better idea to implement it?

cbosdo: why not storing the result of `mgr-setup` in a variable and return it at the end to keep the rm in the script?

mgradm/shared/templates/mgrSetupScriptTemplate.go
```diff
@@ -20,10 +20,12 @@ echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:800
 echo 'JAVA_OPTS=" $JAVA_OPTS -Xdebug -Xrunjdwp:transport=dt_socket,address=*:8002,server=y,suspend=n" ' >> /usr/share/rhn/config-defaults/rhn_search_daemon.conf
 {{- end }}
 
-/usr/lib/susemanager/bin/mgr-setup -s -n
+RESULT=$(/usr/lib/susemanager/bin/mgr-setup -s -n)
```
cbosdo, line 23: Is it the same then
```suggestion
/usr/lib/susemanager/bin/mgr-setup -s -n
RESULT=$?
```

cbosdo (approved):

PR #357, opened 2024-06-11
Improved Hub XML-RPC support

nadvornik:
## What does this PR change?

This PR adds Hub XML-RPC support to podman migrate and upgrade and to all kubernetes commands.

Also, it fixes some kubernetes problems:
- incorrect arg to GetNode calls
- error detection on traefik setup

This PR depends on https://github.com/uyuni-project/uyuni/pull/8902

## Test coverage
- No tests: **add explanation**
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/292

- [ ] **DONE**

## Changelogs

Make sure the changelogs entries you are adding are compliant with https://github.com/uyuni-project/uyuni/wiki/Contributing#changelogs and https://github.com/uyuni-project/uyuni/wiki/Contributing#uyuni-projectuyuni-repository

If you don't need a changelog check, please mark this checkbox:

- [ ] No changelog needed

If you uncheck the checkbox after the PR is created, you will need to re-run `changelog_test` (see below)

# Before you merge

Check [How to branch and merge properly](https://github.com/uyuni-project/uyuni/wiki/How-to-branch-and-merge-properly)!

PR #400, opened 2024-07-12
Inspection refactoring and cleanup

cbosdo:
## What does this PR change?

Split the inspection scripts based on their usage

It doesn't make sense to try to extract all the values for both
containers and hosts. This commit introduces Inspector objects that are
responsible for generating the inspection script and parsing it into a
structure.

While at it, the extracted values that were not used have been removed.
This gains quite some time at each execution since extracting the
SCC registration status was quite long.

While doing this we could also drop the InspectData proxyHost flag as
SCC credentials are extracted for both server and proxy hosts.

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): #

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge


aaannz:
We have rather clean dir structure, how about having `shared/utils/inspector`, `shared/utils/inspector/inspect.go`, `shared/utils/inspector/host.go` and `shared/utils/inspector/server.go`?

cbosdo:
> We have rather clean dir structure, how about having `shared/utils/inspector`, `shared/utils/inspector/inspect.go`, `shared/utils/inspector/host.go` and `shared/utils/inspector/server.go`?

I was hesitating doing it... looks like I'll do it if we have the same idea 😉

mbussolotto (approved):
LGTM

shared/podman/login.go
```diff
@@ -0,0 +1,81 @@
...
+		return authFilePath, func() {
+			os.Remove(authFilePath)
+		}, nil
+	}
+
+	return "", func() {}, nil
+}
+
+// InspectHost check values on a host machine.
+func inspectHost() (*hostInspectData, error) {
```
rjmateus, line 59: Since this is related to the host impector, I would move this to the file `hostinspector` in the same directory. It uses the structure from there, and this way it would be all concentrated in the same file

mgradm/cmd/migrate/podman/utils.go
```diff
@@ -82,6 +83,21 @@ func migrateToPodman(globalFlags *types.GlobalFlags, flags *podmanMigrateFlags,
 		return err
 	}
 
+	// Prepare confidential computing containers
```
rjmateus, line 86: Should the prepare for XML-RPC container also be added in here? Or will it by handled by the other pr?

rjmateus (approved):
lgtm, just a few nitpicks

shared/podman/login.go
cbosdo: Oops! I did it in my original branch but forgot to move it in this one.

mgradm/cmd/migrate/podman/utils.go
cbosdo: It's part of https://github.com/uyuni-project/uyuni-tools/pull/357 but I wonder if I should start merging all those bits into one PR to avoid missing some

shared/podman/login.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	if creds.SccPassword != "" && creds.SccUsername != "" {
+		// We have SCC credentials, so we are pretty likely to need registry.suse.com
+		token := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", creds.SccUsername, creds.SccPassword)))
+		authFileContent := fmt.Sprintf(`{
+	"auths": {
+		"registry.suse.com" : {
```
rjmateus, line 28: Can we hard code this registry when we can have a different registry name on the cloud?

cbosdo: This is only set if we have SCC credentials... so no other registry uses them

PR #402, opened 2024-07-15
Check if volume directory is mounted when purging

aaannz:
## What does this PR change?

Mounted volume cannot be removed and podman thus exits with non- zero return even though data and internal record were purged. This commit adds mount check on podman data removal and reports error only when volume dir is not empty and not mounted.

This commit also fixes error handling when volume purge was not successful.

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24417

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/podman/utils.go
+++ b/shared/podman/utils.go
@@ -5,8 +5,11 @@
 package podman
 
 import (
+	"errors"
+	"io"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 
 	"github.com/rs/zerolog"
@@ -155,9 +158,16 @@ func DeleteVolume(name string, dryRun bool) error {
 			log.Info().Msgf(L("Would run %s"), "podman volume rm "+name)
 		} else {
 			log.Info().Msgf(L("Run %s"), "podman volume rm "+name)
-			err := utils.RunCmd("podman", "volume", "rm", name)
-			if err != nil {
-				log.Error().Err(err).Msgf(L("Failed to remove volume %s"), name)
+			if err := utils.RunCmd("podman", "volume", "rm", name); err != nil {
+				// Check if the volume is not mounted - for example var-pgsql as second storage device
+				// We cannot get volume real path from volume inspect because podman already removed it
+				// from its internal db. Let's assume our default path
+				target := path.Join("/var/lib/containers/storage/volumes", name)
+				if isVolumePathEmpty(target) && isVolumePathMounted(target) {
+					log.Info().Msgf(L("Volume %s is externally mounted, directory cannot be removed"), name)
+					return nil
+				}
+				return err
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


shared/podman/utils.go
```diff
@@ -155,21 +158,51 @@ func DeleteVolume(name string, dryRun bool) error {
...
 		} else {
 			log.Info().Msgf(L("Run %s"), "podman volume rm "+name)
-			err := utils.RunCmd("podman", "volume", "rm", name)
-			if err != nil {
-				log.Error().Err(err).Msgf(L("Failed to remove volume %s"), name)
+			if err := utils.RunCmd("podman", "volume", "rm", name); err != nil {
+				// Check if the volume is not mounted - for example var-pgsql as second storage device
+				// We cannot get volume real path from volume inspect because podman already removed it
+				// from its internal db. Let's assume our default path
+				target := path.Join("/var/lib/containers/storage/volumes", name)
```
cbosdo, line 165: Can't we at least get the storage root from `podman system info --format '{{ .Store.GraphRoot }}'` instead of assuming a default that can be changed in a config file?

shared/podman/utils.go
```diff
@@ -155,21 +158,51 @@ func DeleteVolume(name string, dryRun bool) error {
...
 	}
-	return cmd.ProcessState.ExitCode() == 0
+	return cmd.ProcessState.Success()
+}
+
+func isVolumePathMounted(volume string) bool {
+	cmd := exec.Command("findmnt", "--target", volume)
+	var exitError *exec.ExitError
+	if err := cmd.Run(); err != nil && errors.As(err, &exitError) {
+		log.Debug().Err(err).Msgf("findmnt --target %s", volume)
```
cbosdo, line 191: Will `findmnt` be available on all recent enough distros where Uyuni could run?

shared/podman/utils.go
aaannz: That's a good point. Will do.

rjmateus (approved):
Looks good to me. I would only change what Cedric mentioned about the podman volumes location

shared/podman/utils.go
cbosdo: Let's keep it as is for now and change if someone complains

shared/podman/utils.go
aaannz: Done

shared/podman/utils.go
```diff
@@ -155,21 +158,62 @@ func DeleteVolume(name string, dryRun bool) error {
...
+	defer f.Close()
+
+	_, err = f.Readdirnames(1)
+	return err == io.EOF
+}
+
+func getPodmanVolumeBasePath() (string, error) {
+	cmd := exec.Command("podman", "system", "info", "--format={{ .Store.VolumePath }}")
+	out, err := cmd.Output()
+	return strings.TrimSuffix(string(out), "\n"), err
```
cbosdo, line 216: I think `TrimSpaces` is good enough for this:

```suggestion
	return strings.TrimSpaces(string(out)), err
```

cbosdo (approved):
nitpick about trimming function, but all is good to me

Issue #239, opened 2024-04-12
Add `mgrctl connect` command to store the server API credentials in configuration file


cbosdo:
To avoid passing the API credentials for every call, adding an `mgrctl connect` command testing the connection to the API server and storing the credentials in the config file would be cool.

mmmmmmpc:
Very convenient for the user, indeed. 
It will make the behaviour similar to `kubectl`, something becoming a well known control tool.

aaannz:
PRs merged, all done here

PR #410, opened 2024-07-19
Add support for login and logout api options

aaannz:
## What does this PR change?

Add support for login and logout api options.

Login will store session cookie together with server to the ~/.config/.uyuni-api file.
Next api calls will try to load session from this file and use it.

If session is detected to be expired (by issuing get user/listAssignableRoles), if mgrctl api is run in terminal it asks for user and password. If not, call is aborted.

Logout command removes the file with stored session and also calls auth/logout.

## Test coverage
- Unit tests were added

- [X] **DONE**

## Links

Issue(s): https://github.com/uyuni-project/uyuni-tools/issues/239
Port: https://github.com/SUSE/uyuni-tools/pull/14

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -213,6 +213,8 @@ golang.org/x/crypto v0.0.0-20181029021203-45a5f77698d3/go.mod h1:6SG95UA2DQfeDnf
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/crypto v0.0.0-20190510104115-cbcb75029529/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
 golang.org/x/crypto v0.0.0-20190605123033-f99c8df09eb5/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
+golang.org/x/crypto v0.25.0 h1:ypSNr+bnYL2YhwoMt2zPxHFmbAN1KZs/njMG3hxUp30=
+golang.org/x/crypto v0.25.0/go.mod h1:T+wALwcMOSE0kXgUAnPAHqTLW+XHgcELELW8VaDgm/M=
 golang.org/x/exp v0.0.0-20190121172915-509febef88a4/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190306152737-a1d7652674e8/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190510132918-efd6b22b2522/go.mod h1:ZjyILWgesfNpC6sMxTJOJm9Kp84zZh5NQWvqDGG3Qr8=
@@ -272,12 +274,18 @@ golang.org/x/sys v0.0.0-20220811171246-fbc7d0a398ab/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.12.0 h1:CM0HF96J0hcLAwsHPJZjfdNzs0gftsLfgKt57wWHJ0o=
 golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.22.0 h1:RI27ohtqKCnwULzJLqkv897zojh5/DwS/ENaMzUOaWI=
+golang.org/x/sys v0.22.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/term v0.10.0 h1:3R7pNqamzBraeqj/Tj8qt1aQ2HpmlC+Cx/qL/7hn4/c=
 golang.org/x/term v0.10.0/go.mod h1:lpqdcUyK/oCiQxvxVrppt5ggO2KCZ5QblwqPnfZ6d5o=
+golang.org/x/term v0.22.0 h1:BbsgPEJULsl2fV/AT3v15Mjva5yXKQDyKf+TbDz7QJk=
+golang.org/x/term v0.22.0/go.mod h1:F3qCibpT5AMpCRfhfT53vVJwhLtIVHhB9XDjfFvnMI4=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/text v0.3.1-0.20180807135948-17ff2d5776d2/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/text v0.3.2 h1:tW2bmiBqwgJj/UpqtC8EpXEZVYOwU0yG4iWbprSVAcs=
 golang.org/x/text v0.3.2/go.mod h1:bEr9sfX3Q8Zfm5fL9x+3itogRgK3+ptLWKqgva+5dAk=
+golang.org/x/text v0.16.0 h1:a94ExnEXNtEwYLGJSIUxnWoxoRz/ZcCsV63ROupILh4=
+golang.org/x/text v0.16.0/go.mod h1:GhwF1Be+LQoKShO3cGOHzqOgRrGaYc9AvblQOmPVHnI=
 golang.org/x/time v0.0.0-20181108054448-85acf8d2951c/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
 golang.org/x/time v0.0.0-20190308202827-9d24e82272b4/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
 golang.org/x/tools v0.0.0-20180221164845-07fd8470d635/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
--- a/go.sum
+++ b/go.sum
@@ -272,17 +272,12 @@ golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.0.0-20210927094055-39ccf1dd6fa6/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.0.0-20220811171246-fbc7d0a398ab/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
-golang.org/x/sys v0.12.0 h1:CM0HF96J0hcLAwsHPJZjfdNzs0gftsLfgKt57wWHJ0o=
-golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.22.0 h1:RI27ohtqKCnwULzJLqkv897zojh5/DwS/ENaMzUOaWI=
 golang.org/x/sys v0.22.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
-golang.org/x/term v0.10.0 h1:3R7pNqamzBraeqj/Tj8qt1aQ2HpmlC+Cx/qL/7hn4/c=
-golang.org/x/term v0.10.0/go.mod h1:lpqdcUyK/oCiQxvxVrppt5ggO2KCZ5QblwqPnfZ6d5o=
 golang.org/x/term v0.22.0 h1:BbsgPEJULsl2fV/AT3v15Mjva5yXKQDyKf+TbDz7QJk=
 golang.org/x/term v0.22.0/go.mod h1:F3qCibpT5AMpCRfhfT53vVJwhLtIVHhB9XDjfFvnMI4=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/text v0.3.1-0.20180807135948-17ff2d5776d2/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
-golang.org/x/text v0.3.2 h1:tW2bmiBqwgJj/UpqtC8EpXEZVYOwU0yG4iWbprSVAcs=
 golang.org/x/text v0.3.2/go.mod h1:bEr9sfX3Q8Zfm5fL9x+3itogRgK3+ptLWKqgva+5dAk=
 golang.org/x/text v0.16.0 h1:a94ExnEXNtEwYLGJSIUxnWoxoRz/ZcCsV63ROupILh4=
 golang.org/x/text v0.16.0/go.mod h1:GhwF1Be+LQoKShO3cGOHzqOgRrGaYc9AvblQOmPVHnI=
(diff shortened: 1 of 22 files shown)


go.sum
```diff
@@ -213,6 +213,8 @@ golang.org/x/crypto v0.0.0-20181029021203-45a5f77698d3/go.mod h1:6SG95UA2DQfeDnf
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/crypto v0.0.0-20190510104115-cbcb75029529/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
 golang.org/x/crypto v0.0.0-20190605123033-f99c8df09eb5/go.mod h1:yigFU9vqHzYiE8UmvKecakEJjdnWj3jj499lnFckfCI=
+golang.org/x/crypto v0.25.0 h1:ypSNr+bnYL2YhwoMt2zPxHFmbAN1KZs/njMG3hxUp30=
+golang.org/x/crypto v0.25.0/go.mod h1:T+wALwcMOSE0kXgUAnPAHqTLW+XHgcELELW8VaDgm/M=
```
cbosdo, line 217: Would there be a way to only have one version of each module we use?

go.sum
```diff
@@ -272,12 +274,18 @@ golang.org/x/sys v0.0.0-20220811171246-fbc7d0a398ab/go.mod h1:oPkhp1MJrh7nUepCBc
 golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.12.0 h1:CM0HF96J0hcLAwsHPJZjfdNzs0gftsLfgKt57wWHJ0o=
 golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.22.0 h1:RI27ohtqKCnwULzJLqkv897zojh5/DwS/ENaMzUOaWI=
+golang.org/x/sys v0.22.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
```
cbosdo, line 278: we start having a gazillion versions of it...

mgrctl/cmd/api/api.go
```diff
@@ -43,11 +43,29 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
 		},
 	}
 
+	apiLogin := &cobra.Command{
+		Use:   "login",
+		Short: L("Store login information for future API usage"),
+		Long:  L("Login stores login information for next API calls. User name, password and remote host can be provided using flag or will be asked interactively. Environmental variables are also supported."),
```
cbosdo, line 49: may be you should split this long line: I don't think those are wrapped in the terminal.

mgrctl/cmd/api/login.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/uyuni-project/uyuni-tools/shared/types"
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
+)
+
+func runLogin(globalFlags *types.GlobalFlags, flags *apiFlags, cmd *cobra.Command, args []string) error {
+	log.Debug().Msg("Running login command")
+
+	utils.AskIfMissing(&flags.User, cmd.Flag("api-user").Usage, 0, 0, nil)
+	utils.AskPasswordIfMissing(&flags.Password, cmd.Flag("api-password").Usage, 0, 0)
+	// ToDO add FQDN checker from rebase
```
cbosdo, line 22:
```suggestion
	// TODO add FQDN checker from rebase
```
Most editors highlight `TODO` comments all caps, not sure about mixed casing

mgrctl/cmd/api/login.go
```diff
@@ -0,0 +1,48 @@
...
+	utils.AskIfMissing(&flags.Server, cmd.Flag("api-server").Usage, 0, 0, nil)
+
+	if err := api.StoreLoginCreds(&flags.ConnectionDetails); err != nil {
```
cbosdo, line 25: Don't you want to check for existing stored credentials before running the login?

shared/api/credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return err
+	}
+
+	// store encrypted credentials into separate credentials config storage
+	auth := authStorage{
+		User:     encUser,
+		Password: encPassword,
+		Server:   connection.Server,
+	}
```
cbosdo, line 43: Shouldn't we start storing a list of credentials? In the future I can imagine `mgrctl` installed on a laptop and the user would use it to query to multiple Uyuni servers (hub scenario or test/dev/prod setup).

I don't mean to change the functions dealing with credentials, but only the storage file to make it future-proof.

shared/api/credentials.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func getAPICredsFile() string {
+	xdgConfigHome := os.Getenv("XDG_CONFIG_HOME")
+	if xdgConfigHome == "" {
+		home, err := os.UserHomeDir()
+		if err == nil {
+			xdgConfigHome = path.Join(home, ".config")
+		}
+	}
```
cbosdo, line 133: We have another place where we compute the user home dir, in the configuration code. May be those could be shared.

shared/api/credentials.go
```diff
@@ -0,0 +1,167 @@
...
+	return res
+}
+
+func encryptMsg(message string, secret [32]byte) ([]byte, error) {
+	var nonce [24]byte
+	if _, err := io.ReadFull(rand.Reader, nonce[:]); err != nil {
+		return nil, err
+	}
+
+	return secretbox.Seal(nonce[:], []byte(message), &nonce, &secret), nil
```
cbosdo, line 156: It's a false sense of security here since the key is the server URL and it's stored in clear text. Is it worth the pain? or just a plain text file with `0600` permissions would be enough?

go.sum
aaannz: Indeed, I will need to check how to tackle this. So far this was all handled by vscode and `go get`

mgrctl/cmd/api/login.go
aaannz: Ah, I forgot about this. Now with all PRs merged and want to use https://github.com/uyuni-project/uyuni-tools/blob/main/shared/utils/utils.go#L411

mgrctl/cmd/api/login.go
aaannz: Currently I indeed just overwrite what's stored. I'll add a check and `--force|-f` option.

shared/api/credentials.go
aaannz: Maybe change it to

```go
authStorage{
  $server: {
    User,
    Password,
  }
}
```
?

shared/api/credentials.go
aaannz: I was looking into this, but currently did not find something which is stable and sort of secret. I tried tpm functions, but then not everyone has tpm in working order.

Currently it indeed protects nothing, but we can discuss what to use as a secret and what we want to protect against.

IMO to prevent against credentials leak we would need to use hw tokens or something like that. Or maybe a URL + date (YYYYMMDD) so login would be valid only for a day?

To prevent accidental disclosure we might read something not included in support config, but specific to system and stable. Like `/var/lib/zypp/AnonymousUniqueId` for example. But that is available only on SUSE systems. Or generate some random string and store it somewhere.

Overall I wanted to have this discussion in the PR or during the MU2 devel time.

shared/api/credentials.go
cbosdo: Either this or using an array as you prefer.
```
authStorage: [
  {
     Server,
     User,
     Password,
  }
]
```

shared/api/credentials.go
cbosdo: Since spacewalk and the others are storing the credentials as plain text in a config file and only rely on permissions, can we also do this? @rjmateus @admd any take on this?

aaannz: We can be a little better then 10years old software 😄 
For example `spacecmd` is not storing password. It stores user name and session id, which I think is specific to xmlrpc.

Generating some random thing and storing it somewhere is I thing least we can do and can help if credentials file is accidentally disclosed.

aaannz: At the end I found out that I am stupid, again.

I was so focused on credentials storage and how to do that that I forgot that we have pxt-session-cookie. Even writing above comment did not trigger me.

So I rewrote it almost completely. Removed that crypto thingie, now we just store pxt-session-cookie like spacecmd does.

go.sum
aaannz: At the end I did not need any new dependency, now this only remains as and dependency update. Should I remove it?

go.sum
aaannz: `go mod tidy` seems to help. Applied.

go.sum
cbosdo: No, keep it if `go mod tidy` keeps it

mgrctl/cmd/api/api.go
```diff
@@ -28,7 +29,12 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
 	apiGet := &cobra.Command{
 		Use:   "get path [parameters]...",
 		Short: L("Call API GET request"),
-		Long:  L("Takes an API path and optional parameters and then issues GET request with them. If user and password are provided, calls login before API call"),
+		Long: L(`Takes an API path and optional parameters and then issues GET request with them.
+
+If user and password are provided, calls login before API call
+
+Example:
+# mgrctl api get user/getDetails login=test`),
```
cbosdo, line 37: It's really nice to have added an example! I'm sure it'll be very useful to users

mgrctl/cmd/api/api.go
```diff
@@ -37,17 +43,48 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
...
+		Long: L(`Takes an API path and parameters and then issues POST request with them.
+User and password are mandatory.
+
+Parameters can be either JSON encoded string or one or more key=value pairs.
+
+Key=Value pairs example:
+# mgrctl api post user/create login=test password=testXX firstName=F lastName=L email=test@localhost
+
+JSON example:
+# mgrctl api post user/create '{"login":"test", "password":"testXX", "firstName":"F", "lastName":"L", "email":"test@localhost"}'`),
```
cbosdo, line 55: Also cool to have added both kinds of examples, thanks!

mgrctl/cmd/api/api.go
```diff
@@ -37,17 +43,48 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
 	apiPost := &cobra.Command{
 		Use:   "post path parameters...",
 		Short: L("Call API POST request"),
-		Long:  L("Takes an API path and parameters and then issues POST request with them. User and password are mandatory. Parameters can be either JSON encoded string or one or more key=value pairs."),
+		Long: L(`Takes an API path and parameters and then issues POST request with them.
+User and password are mandatory.
```
cbosdo, line 47: Are the user and password really mandatory if the login was done before? Shouldn't this sentence be aligned with the similar one for `get`?

mgrctl/cmd/api/api.go
```diff
@@ -37,17 +43,48 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
...
 		},
 	}
 
+	apiLogin := &cobra.Command{
+		Use:   "login",
+		Short: L("Store login information for future API usage"),
+		Long: L(`Login stores login information for next API calls.
+
+User name, password and remote host can be provided using flags or will be asked interactively.
+Environmental variables are also supported.`),
```
cbosdo, line 67:
```suggestion
Environment variables are also supported.`),
```

mgrctl/cmd/api/api.go
```diff
@@ -37,17 +43,48 @@ func NewCommand(globalFlags *types.GlobalFlags) (*cobra.Command, error) {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return utils.CommandHelper(globalFlags, cmd, args, &flags, runLogin)
+		},
+	}
+	apiLogin.Flags().BoolP("force", "f", false, L("Overwrite existing login if exists"))
```
cbosdo, line 72: can't people just call `mgrctl logout` before instead of having `--force` ? `podman login` doesn't have such a parameter.

mgrctl/cmd/api/login.go
```diff
@@ -0,0 +1,53 @@
...
+	utils.AskIfMissing(&flags.Server, cmd.Flag("api-server").Usage, 0, 0, utils.IsWellFormedFQDN)
+	utils.AskIfMissing(&flags.User, cmd.Flag("api-user").Usage, 0, 0, nil)
+	utils.AskPasswordIfMissing(&flags.Password, cmd.Flag("api-password").Usage, 0, 0)
+
+	client, err := api.Init(&flags.ConnectionDetails)
+	if err != nil {
+		return err
+	}
+	if err := client.Login(); err != nil {
+		return utils.Errorf(err, L("Failed to validate credentials. Not storing"))
```
cbosdo, line 35:
```suggestion
		return utils.Errorf(err, L("Failed to validate credentials."))
```

No need to say the invalid credentials won't be stored: this is an expected behavior.

mgrctl/cmd/api/login.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	if err := api.StoreLoginCreds(client); err != nil {
+		return err
+	}
+
+	log.Info().Msg(L("Login credentials verified and stored"))
```
cbosdo, line 41:
```suggestion
	log.Info().Msgf(L("Logged in to %s"), flags.Server)
```

Same here, verified credentials are expected to be stored, no need to log it.

shared/api/api.go
```diff
@@ -5,122 +5,73 @@
...
-		}
-	}
-	return nil
 }
 
-func prettyPrint(v interface{}) string {
+func logTraceHeader(v *http.Header) {
+	if log.Logger.GetLevel() != zerolog.TraceLevel {
+		return
+	}
```
cbosdo, line 38: Isn't `log.Trace()` enough? or did you add this to avoid useless json marshalling? If so, a comment could be nice.

shared/api/api.go
```diff
@@ -5,122 +5,73 @@
...
 
 	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusBadRequest {
 		var errResponse map[string]string
-		if err = json.NewDecoder(res.Body).Decode(&errResponse); err == nil {
-			return nil, fmt.Errorf(errResponse["message"])
+		if res.Body != nil {
+			body, err := io.ReadAll(res.Body)
+			if err == nil {
+				if err = json.Unmarshal(body, &errResponse); err == nil {
+					return nil, fmt.Errorf(errResponse["message"])
```
cbosdo, line 70: Shouldn't this be for `err != nil` ?

shared/api/credentials.go
```diff
@@ -0,0 +1,125 @@
...
+func StoreLoginCreds(client *APIClient) error {
+	if client.AuthCookie.Value == "" {
+		return fmt.Errorf(L("not logged in, session cookie is missing"))
+	}
+	// Future: Add support for more servers if needed in the future
+	auth := []authStorage{
+		{
+			Session: client.AuthCookie.Value,
+			Server:  client.Details.Server,
+			CACert:  client.Details.CAcert,
```
cbosdo, line 29: Do we really want to store the CA cert here? Isn't that the job of the system SSL setup?

shared/api/credentials.go
```diff
@@ -0,0 +1,125 @@
...
+		return utils.Errorf(err, L("unable to read credentials file %s"), getAPICredsFile())
+	}
+	authStore := []authStorage{}
+	err = json.Unmarshal(data, &authStore)
+	if err != nil {
+		return utils.Errorf(err, L("unable to decode credentials file"))
+	}
+
+	if len(authStore) == 0 {
+		return fmt.Errorf(L("no credentials loaded"))
```
cbosdo, line 96: Is this really an error? I suppose this is expected at first login

shared/api/api.go
aaannz: Indeed, I added it to prevent useless marshaling.

shared/api/api.go
aaannz: Nope, this is when we are able to unmarshal error response. I can switch those branches to get expected `!=`

shared/api/credentials.go
aaannz: Agree that `CACert` name is misleading. This is only path to the certificate which we want to remember for next call. I will rename it to be self explaining.

cbosdo: which certificate? The server one?

shared/api/api.go
cbosdo: Then maybe add a comment before that check to state this: I suspect this will be overlooked in the future and could go away in a too zealous cleanup

shared/api/credentials.go
aaannz: CA which signed server certificate.

aaannz: Renamed to `CApath` so it is clear what is expected here.

shared/api/credentials.go
aaannz: This only happen if file with stored session exists, but something is not working correctly.

shared/api/api.go
aaannz: Comment added

mgrctl/cmd/api/login.go
```diff
@@ -0,0 +1,53 @@
...
+	. "github.com/uyuni-project/uyuni-tools/shared/l10n"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
+	"github.com/uyuni-project/uyuni-tools/shared/utils"
+)
+
+func runLogin(globalFlags *types.GlobalFlags, flags *apiFlags, cmd *cobra.Command, args []string) error {
+	log.Debug().Msg("Running login command")
+
+	if api.IsAlreadyLoggedIn() && !flags.ForceLogin {
+		return fmt.Errorf(L("Refusing to overwrite existing login. Use --force to ignore this check."))
```
cbosdo, line 23:
```suggestion
		return fmt.Errorf(L("Refusing to overwrite existing login. Use --force-login to ignore this check."))
```

cbosdo: oops, indeed. I'm not sure `--force` would indicate that it's about login, so yes something like `--force-login` or `--forcelogin` would make it more explicit for users.

PR #411, opened 2024-07-22
migration fails when old tomcat configuration files are present in the source system (bsc#1228183)

mbussolotto:
## What does this PR change?

see title 
## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24877
Issue(s): https://github.com/uyuni-project/uyuni/issues/8946

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/templates/migrateScriptTemplate.go
+++ b/mgradm/shared/templates/migrateScriptTemplate.go
@@ -48,6 +48,10 @@ rpm -qa --qf '[%{fileflags},%{filenames}\n]' |grep ",/etc/" | while IFS="," read
     fi
 done
 
+# exclude tomcat default configuration. All settings should be store in /etc/tomcat/conf.d/
+echo "-/ /etc/sysconfig/tomcat" >> exclude_list
+echo "-/ /etc/tomcat/tomcat.conf" >> exclude_list
+
 # exclude schema migration files
 echo "-/ /etc/sysconfig/rhn/reportdb-schema-upgrade" >> exclude_list
 echo "-/ /etc/sysconfig/rhn/schema-upgrade" >> exclude_list
@@ -80,6 +84,14 @@ while IFS="," read -r target path ; do
   fi
 done < distros
 
+if $SSH {{ .SourceFqdn }} test -e /etc/tomcat/conf.d; then
+  echo "Copying tomcat configuration.."
+  mkdir -p /etc/tomcat/conf.d
+  rsync -e "$SSH" --rsync-path='sudo rsync' -avz {{ .SourceFqdn }}:/etc/tomcat/conf.d /etc/tomcat/;
+else
+  echo "Skipping tomcat configuration.."
+fi
+
 rm -f /srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT;
 ln -s /etc/pki/trust/anchors/LOCAL-RHN-ORG-TRUSTED-SSL-CERT /srv/www/htdocs/pub/RHN-ORG-TRUSTED-SSL-CERT;
 
@@ -105,10 +117,9 @@ sed 's/client_use_localhost: false/client_use_localhost: true/' -i /etc/cobbler/
 echo "Altering configuration for container environment..."
 sed 's/address=[^:]*:/address=*:/' -i /etc/rhn/taskomatic.conf;
 
-if test ! -f /etc/tomcat/conf.d/remote_debug.conf -a -f /etc/sysconfig/tomcat; then
-  mv /etc/sysconfig/tomcat /etc/tomcat/conf.d/remote_debug.conf
-fi
-
+echo "Altering tomcat configuration..."
+sed 's/--add-modules java.annotation,com.sun.xml.bind://' -i /etc/tomcat/conf.d/*
+sed 's/-XX:-UseConcMarkSweepGC//' -i /etc/tomcat/conf.d/*
 sed 's/address=[^:]*:/address=*:/' -i /etc/tomcat/conf.d/remote_debug.conf
 
 {{ if .Kubernetes }}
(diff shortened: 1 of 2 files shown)


mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -52,6 +52,13 @@ done
 echo "-/ /etc/sysconfig/rhn/reportdb-schema-upgrade" >> exclude_list
 echo "-/ /etc/sysconfig/rhn/schema-upgrade" >> exclude_list
 
+
+# exclude old tomcat configuration files. The settings present in the files
+# below would be provided by the image. The user configuration are expected
+# in files in /etc/tomcat/conf.d/
+echo "-/ /etc/sysconfig/tomcat" >> exclude_list
```
rjmateus, line 59: What will happen if the customer have changed the tomcat memory configuration, as we suggest in the tunning guide? https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html#tomcat-xmx
I think we have 2 options:
1. Copy all the configuration files, even the `tomcat_java_opts_suse.conf` and filter/replace the two parameters
2. Keep it as is on the PR, but add a note to the migration guide advising the customer to tune the server again, and mentioning some properties are not copied.

mbussolotto: mmm maybe I'm missing something. With this PR we're excluding `/etc/sysconfig/tomcat` and `/etc/tomcat/conf.d/remote_debug.conf`, so `tomcat_java_opts_suse.conf` would be migrated. The offending parameters should be in the file excluded, not in  `tomcat_java_opts_suse.conf` of other files, so no filter/replacement would be required

mbussolotto: ok, I got your point. Actually the wrong parameters are present also in `tomcat_java_opts.conf` if the user changed it

rjmateus: I miss-read the change, and according to @nadvornik the file `tomcat_java_opts_suse.conf` was not copy already.
The main question is that is customer have installed the suma server before the change to extract those parameters to the `tomcat_java_opts_suse.conf` file, those configurations would land on the file `tomcat_java_opts_suse.conf`

mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -108,7 +117,13 @@ sed 's/address=[^:]*:/address=*:/' -i /etc/rhn/taskomatic.conf;
 if test ! -f /etc/tomcat/conf.d/remote_debug.conf -a -f /etc/sysconfig/tomcat; then
   mv /etc/sysconfig/tomcat /etc/tomcat/conf.d/remote_debug.conf
 fi
+if -f /etc/sysconfig/tomcat; then
+  rm /etc/sysconfig/tomcat
+fi
```
nadvornik, line 122: This will remove it even if it comes from the new container. Do we really want this? Maybe exclude_list would be better.

mgradm/shared/templates/migrateScriptTemplate.go
```diff
@@ -80,6 +80,15 @@ while IFS="," read -r target path ; do
...
 
+do
+  if $SSH {{ .SourceFqdn }} test -e /etc/tomcat/conf.d; then
+    echo "Copying tomcat configuration.."
+    mkdir -p /etc/tomcat/conf.d
+    rsync -e "$SSH" --rsync-path='sudo rsync' -avz {{ .SourceFqdn }}:/etc/tomcat/conf.d /etc/tomcat/;
+  else
+    echo "Skipping tomcat configuration.."
+  fi
+
```
nadvornik, line 91: So we want to copy the original tomcat configuration completely, no matter if it is marked config(noreplace) or not? OK.

rjmateus: That is a good point in here Vladimir. How do you suggest to handle the situation where customers could have turned the tomcat parameters?

mbussolotto: > So we want to copy the original tomcat configuration completely, no matter if it is marked config(noreplace) or not? OK.

I was thinking so. Consider the only conf file provided by us are:
- `tomcat_java_opts.conf`
- `tomcat_java_opts_suse.conf` 
and these are the same on 4.3 and 5.0. All the other files are created by the user and if they had some change that works, better copy them.

What do you think?

mgradm/shared/templates/migrateScriptTemplate.go
mbussolotto: adding it to `exclude_list` means it would not be copied at all...the scenario that I was trying to solve is when the user had changes on `/etc/sysconfig/tomcat` (e.g. the remote debug was set here in the past). Probably it's not required at all

mbussolotto:
@rjmateus @nadvornik I changed my idea several times :) but this is now how it works:
- `/etc/sysconfig/tomcat` and `/etc/tomcat/tomcat.conf` are now in exclude list: this is because currently we're not modifying then (now we're creating `tomcat_java_opts.conf` and `tomcat_java_opts_suse.conf`, all user changes should be in `/etc/tomcat/conf.d`
- content of  `/etc/tomcat/conf.d` is copied but we're removing `--add-modules java.annotation,com.sun.xml.bind`  and `-XX:-UseConcMarkSweepGC` from all files in `/etc/tomcat/conf.d` : in this we're preserving all the changes done by the user (also the one mentioned in the tuning doc)

Probably we can add some info in the release note, mentioning that changes done by the user on `/etc/sysconfig/tomcat` and `/etc/tomcat/tomcat.conf` on the source server, should be ported in `/etc/tomcat/conf.d/*` files.

What do you think?

I tested this change with configuration in  bsc#1228183 and it worked

rjmateus (approved):
the approach looks good to me. If Vladimir also agrees, and don't see any issue lets move forward with it.

nadvornik (approved):
LGTM

PR #415, opened 2024-07-29
mgrpxy: expose port on IPv6 network (bsc#1227951)

mbussolotto:
## What does this PR change?

followup of https://github.com/uyuni-project/uyuni-tools/pull/406. Same changes should be present in `mgrpxy`.

## Test coverage
- No tests
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24848

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgrpxy/shared/templates/pod.go
+++ b/mgrpxy/shared/templates/pod.go
@@ -33,6 +33,9 @@ ExecStartPre=/bin/sh -c '/usr/bin/podman pod create --infra-conmon-pidfile %t/uy
 		--network {{ .Network }} \
         {{- range .Ports }}
         -p {{ .Exposed }}:{{ .Port }}{{ if .Protocol }}/{{ .Protocol }}{{ end }} \
+        {{- if $.IPV6Enabled }}
+	-p [::]:{{ .Exposed }}:{{ .Port }}{{if .Protocol}}/{{ .Protocol }}{{end}} \
+        {{- end }}
         {{- end }}
 		--replace ${PODMAN_EXTRA_ARGS}'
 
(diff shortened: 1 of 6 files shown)


mgrpxy/shared/templates/pod.go
```diff
@@ -33,6 +33,7 @@ ExecStartPre=/bin/sh -c '/usr/bin/podman pod create --infra-conmon-pidfile %t/uy
 		--network {{ .Network }} \
         {{- range .Ports }}
         -p {{ .Exposed }}:{{ .Port }}{{ if .Protocol }}/{{ .Protocol }}{{ end }} \
+	-p [::]:{{ .Exposed }}:{{ .Port }}{{if .Protocol}}/{{ .Protocol }}{{end}} \
```
vzhestkov, line 36: I think `[::]` is making it listening for IPv6 and IPv4 as well (at least it works this way with `salt-broker` directly), so 2 entries could be redundant here, but as @aaannz mentioned maybe it makes sense to check if IPv6 is enabled in the network as it's made in `mgradm` case, but I'm not fully sure here as I'm not deep familiar with `podman` network internals.

mbussolotto:
@aaannz @vzhestkov yes, this would work also if ipv6 is disabled. I followed what we implemented for server and trying to keep it simplest as possible, but enabling it just when `ipv6` is enabled is not complicated. I'll go with this change

vzhestkov:
Hmm, it's really interesting. With `salt-broker` if it's set to listen to `[::]` it makes it responsive for both IPv6 and IPv4, but in case of `conmon` with podman it looks like we need to have it listening to both `[::]` and `0.0.0.0`.

vzhestkov (approved):
:+1:

cbosdo (approved):

PR #416, opened 2024-07-30
use scc flags during podman pull

mbussolotto:
## What does this PR change?

use scc flags during podman pull

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/podman/login.go
+++ b/shared/podman/login.go
@@ -9,15 +9,24 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/rs/zerolog/log"
 	. "github.com/uyuni-project/uyuni-tools/shared/l10n"
+	"github.com/uyuni-project/uyuni-tools/shared/types"
 	"github.com/uyuni-project/uyuni-tools/shared/utils"
 )
 
 // PodmanLogin logs in the registry.suse.com registry if needed and returns an authentication file, a cleanup function and an error.
-func PodmanLogin(hostData *HostInspectData) (string, func(), error) {
-	if hostData.SccPassword != "" && hostData.SccUsername != "" {
+func PodmanLogin(hostData *HostInspectData, scc types.SCCCredentials) (string, func(), error) {
+	scc_user := hostData.SccUsername
+	scc_password := hostData.SccPassword
+	if scc.User != "" && scc.Password != "" {
+		log.Info().Msg(L("SCC credentials parameters will be used. SCC credentials from host will be ignored."))
+		scc_user = scc.User
+		scc_password = scc.Password
+	}
+	if scc_user != "" && scc_password != "" {
 		// We have SCC credentials, so we are pretty likely to need registry.suse.com
-		token := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", hostData.SccUsername, hostData.SccPassword)))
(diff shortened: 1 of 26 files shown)


cbosdo (approved):
And now I already smell the question: "where are `--scc-*` parameters on `mgrpxy`?"... Same for kubernetes: we probably will need to add a secret to use for registry using those

mbussolotto:
> And now I already smell the question: "where are --scc-* parameters on mgrpxy?

@cbosdo since it was a 5 min work, I added the flag also for mgrpxy :) I re-asked for review

shared/podman/login.go
```diff
@@ -9,15 +9,27 @@ import (
...
-	if hostData.SccPassword != "" && hostData.SccUsername != "" {
+func PodmanLogin(hostData *HostInspectData, scc types.SCCCredentials) (string, func(), error) {
+	scc_user := hostData.SccUsername
+	scc_password := hostData.SccPassword
+	if scc.User != "" {
+		log.Info().Msg(L("scc-user flag will be used. SCCCredentials in host machine will be ignored."))
+		scc_user = scc.User
+	}
+	if scc.Password != "" {
+		log.Info().Msg(L("scc-password flag will be used. SCCCredentials in host machine will be ignored."))
```
cbosdo, line 27: I would group this warning with the other. If we have only `--scc-user` set we cannot pick the password from the host as it likely wouldn't match

shared/podman/login.go
```diff
@@ -9,15 +9,27 @@ import (
...
 // PodmanLogin logs in the registry.suse.com registry if needed and returns an authentication file, a cleanup function and an error.
-func PodmanLogin(hostData *HostInspectData) (string, func(), error) {
-	if hostData.SccPassword != "" && hostData.SccUsername != "" {
+func PodmanLogin(hostData *HostInspectData, scc types.SCCCredentials) (string, func(), error) {
+	scc_user := hostData.SccUsername
+	scc_password := hostData.SccPassword
+	if scc.User != "" {
+		log.Info().Msg(L("scc-user flag will be used. SCCCredentials in host machine will be ignored."))
+		scc_user = scc.User
+	}
```
cbosdo, line 25:
```suggestion
	if scc.User != "" && scc.Password != "" {
		log.Info().Msg(L("SCC credentials parameters will be used. SCC credentials from host will be ignored."))
		scc_user = scc.User
		scc_password = scc.Password
	}
```

shared/utils/cmd.go
```diff
@@ -54,6 +54,16 @@ func CommandHelper[T interface{}](
 	return fn(globalFlags, flags, cmd, args)
 }
 
+// AddSCCFlag add SCC flags to a command.
+func AddSCCFlag(cmd *cobra.Command) {
+	cmd.Flags().String("scc-user", "", L("SUSE Customer Center username"))
```
cbosdo, line 59: Now we would miss an important help here for `mgradm`: I believe the help should state that those credentials are used to pull the images if coming from registry.suse.com and that on `mgradm` they also set the SCC credentials in the configuration for the products synchronization.

I wonder where to add that piece of help though as it shouldn't be in `mgrpxy`

mbussolotto: probably better to create two different `AddSCCFlag` function, one for `mgradm` and one in `mgrpxy`, in two different files

PR #435, opened 2024-08-21
Add persistent volume for krb5.conf.d

aaannz:
## What does this PR change?

See title. ⬆️ Related server change in [PR here](https://github.com/uyuni-project/uyuni/pull/9188)

## Test coverage
- No tests: already covered
- Unit tests were added

- [ ] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25033

- [X] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/utils/volumes.go
+++ b/shared/utils/volumes.go
@@ -41,6 +41,7 @@ var EtcServerVolumeMounts = []types.VolumeMount{
 	{MountPath: "/etc/sysconfig", Name: "etc-sysconfig"},
 	{MountPath: "/etc/postfix", Name: "etc-postfix"},
 	{MountPath: "/etc/sssd", Name: "etc-sssd"},
+	{MountPath: "/etc/krb5.conf.d", Name: "etc-krb5"},
 }
 
 // EtcServerVolumes represents volumes used for configuration.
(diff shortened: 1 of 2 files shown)


rjmateus:
This is an exceptional case. Do we want to persist this or should we document how to mount those files on the container?
User can have the files in the host OS file system and enhance the podman arg to mount those files in the default locations

aaannz:
Hmm.. I am already providing [modified krb5.conf](https://github.com/uyuni-project/uyuni/pull/9188/files) because of changed path of `default_keytab_name`. So indeed we don't need this and instead have all under `/etc/rhn/krb.conf.d/` which we already have a volume for (`/etc/rhn/`).
We can also modify `/etc/krb.conf.d` to be a link to the `/etc/rhn/krb.conf.d` during the image building.

So documentation would just need to say not to overwrite shipped krb5.conf and instead use one of above mentioned directories.

WDYT?

rjmateus:
I like that idea more, and avoid adding one more volume to the mix. Since we are the ones adding the file `/etc/krb5.conf` manually during build process, we can even add a warning message at the start of the file telling the user to not edit it and add is changes to `/etc/krb5.conf.d` or even better to `/etc/rhn/krb.conf.d/`
Documentation can solve it, otherwise, if the user restart the container he will find out really fast that the krb5 change are gone.

aaannz:
Agreed. Closing this one and will do everything in https://github.com/uyuni-project/uyuni/pull/9188/ and docs.

PR #448, opened 2024-09-10
save supportconfig in current directory (bsc#1226759)

mbussolotto:
## What does this PR change?

save supportconfig in current directory (bsc#1226759)

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25226

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/shared/connection.go
+++ b/shared/connection.go
@@ -440,7 +440,7 @@ func chooseBackend[F interface{}](
 }
 
 // RunSupportConfig will run supportconfig command on given connection.
-func (cnx *Connection) RunSupportConfig() ([]string, error) {
+func (cnx *Connection) RunSupportConfig(tmpDir string) ([]string, error) {
 	var containerTarball string
 	var files []string
 	extensions := []string{"", ".md5"}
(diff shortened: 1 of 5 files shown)


cbosdo (approved):

shared/connection.go
```diff
@@ -450,12 +450,11 @@ func (cnx *Connection) RunSupportConfig() ([]string, error) {
 	}
 
 	// Copy the generated file locally
-	tmpDir, err := os.MkdirTemp("", "mgradm-*")
+	currentDir, err := os.Getwd()
```
aaannz, line 453: What about:

```go
currentDir, err := os.Getwd()
tmpDir, err := os.MkdirTemp(currentDir, "mgradm-*")
```

So we can avoid creating multiple support config files directly under cwd.
Also, are you sure cwd can be written into?

mbussolotto: > So we can avoid creating multiple support config files directly under cwd.

seems a good idea

> Also, are you sure cwd can be written into?

no, but in that case the error should be handled and it should be clear that writing permission are missing. In any case, I'll test it.

aaannz (approved):
Thanks!

PR #451, opened 2024-09-12
share volumes between containers (bsc#1223142)

mbussolotto:
## What does this PR change?

TBH I'm not sure if it'd fix the problem bsc#1223142, but I couldn't find any way to reproduce it and it looks too sporadic, I'd give it a try with it

from podman doc:
```
Labeling systems like SELinux require that proper labels are placed on volume content mounted into a 
<<container|pod>>. 
Without a label, the security system might prevent the processes running inside the 
<<container|pod>> from using the content. 

By default, Podman does not change the labels set by the OS.
To change a label in the <<container|pod>> context, add either of two suffixes :z or :Z to the volume mount. 
These suffixes tell Podman to relabel file objects on the shared volumes. 
The z option tells Podman that two or more <<containers|pods>> share the volume content. 
As a result, Podman labels the content with a shared content label. 
Shared volume labels allow all containers to read/write content. 
The Z option tells Podman to label the content with a private unshared label Only the current <<container|pod>> can use a private volume. 
Relabeling walks the file system under the volume and changes the label on each file, if the volume has thousands of inodes, this process takes a long time, delaying the start of the <<container|pod>>. 
If the volume was previously relabeled with the z option, Podman is optimized to not relabel a second time. 
If files are moved into the volume, then the labels can be manually change with the chcon -R container_file_t PATH command.
```
## Test coverage
- No tests
- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/24144

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/templates/hubXmlrpcServiceTemplate.go
+++ b/mgradm/shared/templates/hubXmlrpcServiceTemplate.go
@@ -37,7 +37,7 @@ ExecStart=/usr/bin/podman run \
         -p {{ .Exposed }}:{{ .Port }}{{if .Protocol}}/{{ .Protocol }}{{end}} \
         {{- end }}
         {{- range .Volumes }}
-        -v {{ .Name }}:{{ .MountPath }} \
+        -v {{ .Name }}:{{ .MountPath }}:z \
         {{- end }}
 	-e HUB_API_URL \
 	-e HUB_CONNECT_TIMEOUT \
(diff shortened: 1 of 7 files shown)


cbosdo (commented):
I'ld rather have the opinion from @thesp0nge for this.

aaannz:
`:z` should be safe, but there might be some performance penalty when relabeling on container start. Hopefully our startup checks will survive it.

mbussolotto:
@cbosdo , @thesp0nge already provided his opinion a while ago here https://github.com/SUSE/spacewalk/issues/24144#issuecomment-2084674167 ;)

rjmateus (approved):

cbosdo (approved):

PR #497, opened 2024-11-14
Fix IsInstalled function

mbussolotto:
## What does this PR change?

see title

## Test coverage
- No tests

- [x] **DONE**

## Links

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -68,8 +68,17 @@ linters:
     - govet
     - lll
     - misspell
+    - revive
     - staticcheck
     - stylecheck
     - unparam
     - unused
     - whitespace
+issues:
+  exclude-rules:
+    - linters:
+        - revive
+      text: "dot-imports"
+    - linters:
+        - revive
+      text: "stutters"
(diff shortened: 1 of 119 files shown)


.golangci.yml
```diff
@@ -51,6 +51,9 @@ linters-settings:
   govet:
     disable:
       - printf
+  #revive:
+  #  rules:
+  #    - name: unused-parameter
```
mbussolotto, line 56: @cbosdo I added this commented part: the analysis works as expected...but seems this is not the only case where we're not using a parameter. Let me know if it's fine to push that and fix the other cases late

cbosdo: Last attempts I had with adding a revive rule, the others were not applied. We need good testing of `revive` config to make sure showing the unused parameters don't hide something else.

cbosdo (approved):

.golangci.yml
cbosdo: I just fixed and enabled the config: using excludes is better as it doesn't disable the other rules.

cbosdo:
> I wonder if all those `_` changes indicates hidden bugs in our code? Like we are supposed to use it but are not using them.

Those remaining in theory are because the function needs to conform to an expected signature. I found a lot of cases where the parameter could simply be removed.

cbosdo:
> I wonder if all those `_` changes indicates hidden bugs in our code? Like we are supposed to use it but are not using them.

I went over the whole diff again and found two of them that could be removed. Basically all the unused parameters in test mocks and in the command entry points have to be turned into `_`. You can see that some of those unused parameters have been simply removed.

PR #537, opened 2025-02-14
fix quoting of arguments

mcalmer:
## What does this PR change?

Should fix

```
:"/usr/bin/sh: line 13: --set-cname: command not found
```

error

## Codespace

## Test coverage
- No tests: already covered

- [x] **DONE**

## Changelogs

- [x] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/install/podman/ssl.go
+++ b/mgradm/cmd/install/podman/ssl.go
@@ -107,7 +107,7 @@ const sslSetupScript = `
 	echo "Generate apache certificate..."
 	cert_args=""
 	for CERT_CNAME in $CERT_CNAMES; do
-		cert_args=$cert_args --set-cname $CERT_CNAME
+		cert_args="$cert_args --set-cname $CERT_CNAME"
 	done
 
 	rhn-ssl-tool --gen-server --no-rpm --cert-expiration 3650 \
--- a/mgradm/cmd/install/podman/ssl.go
+++ b/mgradm/cmd/install/podman/ssl.go
@@ -98,23 +98,23 @@ const sslSetupScript = `
 	echo "Generating the self-signed SSL CA..."
 	mkdir -p /root/ssl-build
 	rhn-ssl-tool --gen-ca --no-rpm --force --dir /root/ssl-build \
-		--password $CERT_PASS \
-		--set-country $CERT_COUNTRY --set-state $CERT_STATE --set-city $CERT_CITY \
-	    --set-org $CERT_O --set-org-unit $CERT_OU \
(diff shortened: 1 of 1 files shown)


aaannz (approved):
LGTM
Looking at the script and also passing it through shellcheck we need much more quotes. Esp. around `$CERT_PASS`.
Can you handle this in our PR or I will create new one.

mgradm/cmd/install/podman/ssl.go
```diff
@@ -98,23 +98,23 @@ const sslSetupScript = `
...
+		--set-country "$CERT_COUNTRY" --set-state "$CERT_STATE" --set-city "$CERT_CITY" \
+	    --set-org "$CERT_O" --set-org-unit "$CERT_OU" \
+	    --set-common-name "$HOSTNAME" --cert-expiration 3650
 	cp /root/ssl-build/RHN-ORG-TRUSTED-SSL-CERT /ssl/ca.crt
 
 	echo "Generate apache certificate..."
 	cert_args=""
 	for CERT_CNAME in $CERT_CNAMES; do
-		cert_args=$cert_args --set-cname $CERT_CNAME
+		cert_args="$cert_args --set-cname \"$CERT_CNAME\""
```
aaannz, line 110: I am not sure this particular thing works as expected. Shellcheck recommends to use arrays in this case (https://www.shellcheck.net/wiki/SC2089).

But can $CERT_CNAME have a space? I don't think so. So I think having quotes just around the whole line is enough.

See:
```bash
📦[aaannz@tumbleweed ~]$ cat test.sh

X="my test dir"
mkdir "$X"

Y="-la \"$X\""

ls $Y

📦[aaannz@tumbleweed ~]$ bash -x test.sh
+ X='my test dir'
+ mkdir 'my test dir'
+ Y='-la "my test dir"'
+ ls -la '"my' test 'dir"'
ls: cannot access '"my': No such file or directory
ls: cannot access 'test': No such file or directory
ls: cannot access 'dir"': No such file or directory
```

cbosdo (approved):

mgradm/cmd/install/podman/ssl.go
mcalmer: I think you can remove it when you have to touch it again but it may not have an effect here.
Hostnames will not have whitespaces. We iterate over $CERT_CNAMES which split at space.
So $CERT_CNAME will not have whitespace and should work with and without quotes, right?

aaannz: Yes, I think so.

PR #539, opened 2025-02-14
SSL certs generation made more robust

cbosdo:
## What does this PR change?

The rhn-ssl-tool script doesn't use only the part before the first dot
in all cases... Implement the same algorithm in sh to find the folder
where the generated certificate is located.

Also use -e to fail the certificate generation and avoid the error
message being lost in the logs.

## Codespace

## Test coverage
- No tests: shell script and covered by the CI

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/install/podman/ssl.go
+++ b/mgradm/cmd/install/podman/ssl.go
@@ -83,7 +83,8 @@ func generateSSLCertificates(image string, flags *adm_utils.ServerFlags, fqdn st
 	command = append(command, envNames...)
 	command = append(command, image)
 
-	command = append(command, "/usr/bin/sh", "-x", "-c", sslSetupScript)
+	// Fail fast with `-e`.
+	command = append(command, "/usr/bin/sh", "-e", "-c", sslSetupScript)
 
 	if _, err := newRunner("podman", command...).Env(envValues).StdMapping().Exec(); err != nil {
 		return []string{}, cleaner, utils.Error(err, L("SSL certificates generation failed"))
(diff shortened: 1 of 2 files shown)


Bischoff (approved):
thanks

mgradm/cmd/install/podman/ssl.go
```diff
@@ -95,6 +96,27 @@ func generateSSLCertificates(image string, flags *adm_utils.ServerFlags, fqdn st
 }
 
 const sslSetupScript = `
+	getMachineName() {
+	  hostname="$1"
+
+	  hostname=$(echo "$hostname" | sed 's/\*/_star_/g')
```
aaannz, line 102:
```suggestion
	  hostname="${hostname//\*/_star_)"
```

Should work too (assuming we call this script using bash)

```bash
$ X="*.test.*.com"
$ echo ${X//\*/_star_}
_star_.test._star_.com
```

mgradm/cmd/install/podman/ssl.go
```diff
@@ -95,6 +96,27 @@ func generateSSLCertificates(image string, flags *adm_utils.ServerFlags, fqdn st
...
+
+	  IFS=. read -r -a hn <<< "$hostname"
```
aaannz, line 104: Why is this needed? Seems like `hn` array is not used?

aaannz: Or was this an attempt to use `field_count=${#hn[@]}` instead of awk used later?

aaannz (commented):
Couple suggestions and also if you can remove `\"` from line 132. I have a feeling that extra `"` passed will break CNAME parsing.

mgradm/cmd/install/podman/ssl.go
cbosdo: We are calling it with `sh`

mgradm/cmd/install/podman/ssl.go
cbosdo: oh indeed, this is a remnant from the first versions of the script...

PR #560, opened 2025-04-10
Check for restorecon presence before using it

cbosdo:
## What does this PR change?

Host OS without SELinux may not have restorecon tool installed. Check it before using. (bsc#1240698)

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/26902
Ports: https://github.com/SUSE/uyuni-tools/pull/65

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/podman/podman.go
+++ b/mgradm/shared/podman/podman.go
@@ -160,13 +160,15 @@ func RunMigration(
 	}
 
 	// now that everything is migrated, we need to fix SELinux permission
-	for _, volumeMount := range utils.ServerVolumeMounts {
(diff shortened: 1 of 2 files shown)


mbussolotto (approved):
I was thinking something like checking if SELinux is enabled...but probably this solution is valid as well. LGTM

cbosdo:
> I was thinking something like checking if SELinux is enabled...but probably this solution is valid as well. LGTM

You could have SELinux disabled and restorecon installed... I pondered that too and decided the most reliable way is to check for the tool itself.

PR #584, opened 2025-05-19
During migration, improve info about usage on ssl-password and logs(bsc#1243009)

mbussolotto:
## Codespace

## Test coverage
- Unit tests were added

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/27192

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/shared/podman/ssl.go
+++ b/mgradm/shared/podman/ssl.go
@@ -211,7 +211,7 @@ func reuseExistingCertificatesFromMounts(image string, tempDir string, isDatabas
 	rootCA, err := shared_podman.ReadFromContainer(containerName, image, volumes, nil,
 		caCheckPath)
 	if err != nil {
-		log.Debug().Err(err).Msgf("CA file %s not found", caCheckPath)
+		log.Info().Msgf(L("CA file %s not found. New CA and certificates will be created."), caCheckPath)
 		return false, nil
 	}
 
@@ -223,7 +223,7 @@ func reuseExistingCertificatesFromMounts(image string, tempDir string, isDatabas
 	cert, err := shared_podman.ReadFromContainer(containerName, image, volumes, nil,
 		crtCheckPath)
 	if err != nil {
-		log.Debug().Err(err).Msgf("Cert file %s not found", crtCheckPath)
+		log.Info().Msgf(L("Cert file %s not found. A new certificate will be created."), crtCheckPath)
 		return false, nil
 	}
 	if err = os.WriteFile(serverCert, cert, 0444); err != nil {
@@ -234,7 +234,7 @@ func reuseExistingCertificatesFromMounts(image string, tempDir string, isDatabas
 	keyData, err := shared_podman.ReadFromContainer(containerName, image, volumes, nil,
 		keyCheckPath)
 	if err != nil {
-		log.Debug().Err(err).Msgf("Cert Key file %s not found", keyCheckPath)
+		log.Info().Msgf(L("Cert key file %s not found. A new certificate will be created."), keyCheckPath)
 		return false, nil
 	}
 	if err = os.WriteFile(serverKey, keyData, 0400); err != nil {
@@ -302,7 +302,7 @@ func generateServerCertificate(image string, sslFlags *adm_utils.InstallSSLFlags
 		"HOSTNAME":     fqdn,
 	}
 	if err := runSSLContainer(sslSetupServerScript, tempDir, image, tz, env); err != nil {
-		return utils.Error(err, L("Server SSL certificates generation failed"))
+		return utils.Error(err, L("Failed to generate server SSL certificates. Please check the input parameters."))
 	}
 
 	log.Info().Msg(L("Server SSL certificates generated"))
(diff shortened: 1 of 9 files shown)


shared/utils/logUtils_test.go
```diff
@@ -28,6 +28,10 @@ func TestRedact(t *testing.T) {
 			`{"adminLogin":"admin","adminPassword":"secret","email":"no@email.com"}`,
 			`{"adminLogin":"admin","adminPassword":"<REDACTED>","email":"no@email.com"}`,
 		},
+		{
+			`password\n`,
+			`password\n`,
```
cbosdo, line 33: We could add a test like `\"password\": \"foo\"`

shared/ssl/flags.go
```diff
@@ -26,3 +26,19 @@ func AddSSLGenerationFlags(cmd *cobra.Command) {
 	_ = utils.AddFlagToHelpGroupID(cmd, "ssl-org", "ssl")
 	_ = utils.AddFlagToHelpGroupID(cmd, "ssl-ou", "ssl")
 }
+
+func AddSSLDBFlags(cmd *cobra.Command) {
```
cbosdo, line 30: I have a similar change in https://github.com/uyuni-project/uyuni-tools/pull/582/files#diff-e6e4d4727d1321ef331f243286c44b3b63a903f0e6c6a67afed2d26f8fa32263

cbosdo (commented):
I haven't seen where we make the SSL error message a bit more obvious, but may be I'm blind

cbosdo (commented):
I'm not sure we should get the hostname rename change in the same PR as this one: it's a bit risky at this phase of the release

deneb-alpha:
> I'm not sure we should get the hostname rename change in the same PR as this one: it's a bit risky at this phase of the release

I agree. can we split the changes and have the hostname rename in a separate PR?

cbosdo:
> > I'm not sure we should get the hostname rename change in the same PR as this one: it's a bit risky at this phase of the release
> 
> I agree. can we split the changes and have the hostname rename in a separate PR? The rename bsc is even not mentioned in the PR description...

The hostname rename has its own PR, it's just that this one is based on it to reuse some refactoring I made. Let me see if we can just extract the refactoring in a small commit and share it between the two PRs.

mbussolotto:
Just to be more clear, this PR requires changes present in the other PR. Resolving conflict after might be a real mess, so the idea is simply to remove that commit once the other PR will be merged (maybe rebase it will be smart enough).

mgradm/shared/podman/ssl.go
```diff
@@ -211,7 +211,7 @@ func reuseExistingCertificatesFromMounts(image string, tempDir string, isDatabas
 	rootCA, err := shared_podman.ReadFromContainer(containerName, image, volumes, nil,
 		caCheckPath)
 	if err != nil {
-		log.Debug().Err(err).Msgf("CA file %s not found", caCheckPath)
+		log.Info().Msgf(L("CA file %s not found, so a new certificate will be created."), caCheckPath)
```
aaannz, line 214:
```suggestion
		log.Info().Msgf(L("CA file %s not found. New CA and certificates will be created."), caCheckPath)
```

mgradm/shared/utils/cmd_utils.go
```diff
@@ -281,30 +281,13 @@ func AddPgsqlFlags(cmd *cobra.Command) {
...
 	AddImageFlag(cmd)
 	AddSCCFlag(cmd)
 	AddPgsqlFlags(cmd)
 	AddDBFlags(cmd)
 	AddReportDBFlags(cmd)
 	ssl.AddSSLGenerationFlags(cmd)
-
-	// For generated CA and certificate
-	cmd.Flags().String("ssl-password", "", L("Password for the CA key to generate"))
-	_ = utils.AddFlagToHelpGroupID(cmd, "ssl-password", "ssl")
```
aaannz, line 295: I think this should go into `ssl.AddSSLGenerationFlags` instead of copying it to all actions.

shared/utils/logUtils.go
```diff
@@ -19,7 +19,7 @@ import (
 	"gopkg.in/natefinch/lumberjack.v2"
 )
 
-var redactRegex = regexp.MustCompile(`([pP]assword[\t :"\\]+)[^\t "\\]+`)
+var redactRegex = regexp.MustCompile(`([pP]assword[\t :"\\]+)[^\\][^\t "\\]+`)
```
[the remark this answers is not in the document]
mbussolotto: this is what happens with the previous regex, according to the unit tests added:
```
logUtils_test.go:48: Testcase 3: Expected password\n got password\<REDACTED> when redacting  password\n
```

and this cause `- invalid character '<' in string escape code****`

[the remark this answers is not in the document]
aaannz: Oh, so when the input is just `password` without anything, we want redacted output be `password` as well, right?

But wouldn't `password.\n` trigger this problem too?

This does not seem to me like a correct way to fix the escape code. That `<REDACTED>` string is what we add.

mgradm/shared/utils/cmd_utils.go
aaannz: Hmm.. true. With k8s cert-manager we do not need CA password.

Ok, something to solve later.

shared/utils/logUtils.go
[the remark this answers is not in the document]
mbussolotto: > Oh, so when the input is just password without anything, we want redacted output be password as well, right?

exactly. The idea of <REDACT> is to use it for the word after password.

> But wouldn't password.\n trigger this problem too?

I can try to add this unit-test too and see if it fits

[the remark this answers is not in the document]
aaannz: I was wondering why would JSON have a problem with `<` and this is because we actually pass `\<RE...` to it. 🤦🏼 

So we need to filter out `password\` things. In this case it makes sense. Ok

PR #591, opened 2025-05-21
Ensure we don't break Debian12 builds

cbosdo:
## What does this PR change?

For debian 12 we still build with go 1.19 and nok8s tag to avoid the backports. Make sure it still builds.

## Test coverage
- No tests: build checks

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

# Before you merge


mbussolotto (approved):

.github/workflows/debian12_build.yml
```diff
@@ -0,0 +1,53 @@
+# SPDX-FileCopyrightText: 2023 SUSE LLC
```
deneb-alpha, line 1:
```suggestion
# SPDX-FileCopyrightText: 2025 SUSE LLC
```
do you think it could make sense to rename the workflow and call it just `debian_build.yml`? I'm thinking at future needs to keep the test when debian 12 will be EOL, but still we could have similar problems with the new stable. 


apart from this, thanks for the nice touch and care for debian :)

cbosdo: I would not even mention debian: it's basically the lowest go we support building with. However I could not come with a satisfactory name.

deneb-alpha: > I would not even mention debian: it's basically the lowest go we support building with. However I could not come with a satisfactory name.

`legacy_go_build.yml`? yes, we are  facing the hardest problem in IT...  naming :sweat_smile:

deneb-alpha: just saw your force push.  also that looks fine

deneb-alpha (approved):
LGTM :)

PR #695, opened 2026-01-08
feat(mgradm): check for podman binary before installation starts

### LINKED ISSUE #621: `mgradm`/`mgrctl`: Check for `podman` right at the beginning

```bash
1:52PM INF Welcome to mgradm
Confirm the password:
Administrator password:
Confirm the password:
```

mihir-dixit2k27:
Description: Fixes #621

What this PR changes? This PR adds a pre-flight check to the mgradm install podman command. It introduces a PreRunE hook that verifies if the podman binary is available in the system PATH before the installation wizard begins.

Why? Previously, if a user ran mgradm install podman without having Podman installed, the tool would proceed through the configuration steps (asking for passwords, etc.) and only fail late in the process when attempting to execute commands. This change ensures the tool "fails fast" with a clear, actionable error message if the prerequisite container engine is missing.

How I tested it:

Compiler Check: Verified clean compilation with go build ./mgradm/cmd/install/podman/.

Negative Test: Simulated a missing binary environment. The tool correctly aborted with:

Error: podman not found in PATH: please install podman before running this command

Positive Test: Verified that the installation wizard proceeds normally when podman is present.


cbosdo (changes requested):

mgradm/cmd/install/install.go
```diff
@@ -38,6 +41,12 @@ NOTE: installing on a remote podman is not supported yet!
 			}
 			return cobra.MaximumNArgs(1)(cmd, args)
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if _, err := exec.LookPath("podman"); err != nil {
+				return fmt.Errorf("podman not found in PATH: please install podman before running this command")
```
cbosdo, line 46: Errors messages need to be wrapped in `L()` function to be able to translate them.
```suggestion
				return errors.New(L("podman not found in PATH: please install podman before running this command"))
```

mgrpxy/cmd/install/install.go
```diff
@@ -39,6 +42,12 @@ NOTE: for now installing on a remote podman is not supported!
 			}
 			return cobra.MaximumNArgs(1)(cmd, args)
 		},
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if _, err := exec.LookPath("podman"); err != nil {
+				return fmt.Errorf("podman not found in PATH: please install podman before running this command")
```
cbosdo, line 47:
```suggestion
				return errors.New(L("podman not found in PATH: please install podman before running this command"))
```

cbosdo (changes requested):
Also merge the 3 commits into a single one with a proper git commit description (usually what you would write as PR description)

PR #696, opened 2026-01-08
feat(mgradm): check for podman binary before installation starts

### LINKED ISSUE #621: `mgradm`/`mgrctl`: Check for `podman` right at the beginning

mihir-dixit2k27:
Fixes #621

What this PR changes? This PR adds a pre-flight check to the mgradm install podman command. It introduces a PreRunE hook that verifies if the podman binary is available in the system PATH before the installation wizard begins.

Why? Previously, if a user ran mgradm install podman without having Podman installed, the tool would proceed through the configuration steps (asking for passwords, etc.) and only fail late in the process when attempting to execute commands. This change ensures the tool "fails fast" with a clear, actionable error message if the prerequisite container engine is missing.

How I tested it:

Compiler Check: Verified clean compilation with go build ./mgradm/cmd/install/podman/.

Negative Test: Simulated a missing binary environment. The tool correctly aborted with:

Error: podman not found in PATH: please install podman before running this command

Positive Test: Verified that the installation wizard proceeds normally when podman is present.

### DIFF
--- a/mgradm/shared/podman/ssl.go
+++ b/mgradm/shared/podman/ssl.go
@@ -274,12 +274,10 @@ func isFQDNMatchingCertificate(fqdn string, certPath string, caPath string) bool
 func reuseExistingCertificatesFromSecrets(isDatabaseCheck bool) bool {
 	if isDatabaseCheck {
 		return shared_podman.HasSecret(shared_podman.DBCASecret) &&
-			shared_podman.HasSecret(shared_podman.DBSSLCertSecret) &&
-			shared_podman.HasSecret(shared_podman.DBSSLKeySecret)
+			shared_podman.HasSecret(shared_podman.DBSSLCertSecret)
 	}
 	return shared_podman.HasSecret(shared_podman.CASecret) &&
-		shared_podman.HasSecret(shared_podman.SSLCertSecret) &&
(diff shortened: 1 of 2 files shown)


rjmateus (commented):
could you squash the commits and add a changes log please? Other then that, looks good to me.

cbosdo (changes requested):

cbosdo (approved):

mihir-dixit2k27:
I have fixed the gofmt formatting issues. The linter should be happy now!

Regarding the commit history: I am having trouble squashing locally due to GPG signing restrictions in my environment. Could you please perform a 'Squash and Merge' when you merge this? Thank you

cbosdo:
> I have fixed the gofmt formatting issues. The linter should be happy now!

I just restarted the workflows to see them green.

> Regarding the commit history: I am having trouble squashing locally due to GPG signing restrictions in my environment. Could you please perform a 'Squash and Merge' when you merge this? Thank you

Do you know that you could just run a `git rebase -i upstream/main` and aggregate the commits there by changing `pick` into `f`? I wonder what restrictions could prevent a rebase…

PR #704, opened 2026-01-17
feat: add podman binary check and separate server image tags

### LINKED ISSUE #621: `mgradm`/`mgrctl`: Check for `podman` right at the beginning

### LINKED ISSUE #673: Separate the server tag from the global one
The `--tag` is used as both a global tag and the server tag. This is cumbersome when the user only wants to change the server tag from the default.
A new `--server-tag` parameter needs to be added for the server container image tag.
For consistency, the `--image` parameter should be renamed `--server-image`.

mihir-dixit2k27:
### Description
This PR combines two improvements for the Podman installer to resolve outstanding issues.

**1. Podman Binary Check (Fixes #621)**
- Implements a pre-flight check to ensure the `podman` binary is present before attempting installation.
- Prevents the installer from failing halfway through due to missing dependencies.

**2. Server Tag Separation (Fixes #673)**
- Renamed `--image` to `--server-image` for clarity.
- Introduced `--server-tag` to allow independent tagging of the server container.
- Updated logic: Falls back to the global `--tag` if `--server-tag` is not specified, preserving backward compatibility.

### Note on Commit History
This branch includes the commit history for the Podman binary check (previously reviewed) alongside the new server tag changes. They are bundled here to ensure all commits retain their verified signatures.

### DIFF
--- a/shared/podman/utils.go
+++ b/shared/podman/utils.go
@@ -51,7 +51,9 @@ var ProxyContainerNames = []string{
 
 // PodmanFlags stores the podman arguments.
 type PodmanFlags struct {
-	Args []string `mapstructure:"arg"`
+    Args []string `mapstructure:"arg"`
+    Name string   `mapstructure:"server-image"`
+    Tag  string   `mapstructure:"server-tag"`
 }
 
 // GetCommonParams splits the common arguments.


cbosdo:
Oops, I merged a PR with a huge refactoring affecting yours. The `mgradm/cmd/install/podman/podman.go` code has been merged to `mgradm/cmd/install/install.go`.
You could surely take this opportunity to run a `git rebase -i upstream/main`, squash some commits and edit those with changes. I could do it here, but it would be my signature instead of yours.

PR #706, opened 2026-01-19
feat: add  separate server image tags ( #673)

### LINKED ISSUE #673: Separate the server tag from the global one

mihir-dixit2k27:
Fixes #673

## Problem

The `--image` and `--tag` flags apply to all containers equally, making it impossible to pin the server container to a specific image or tag without affecting other containers (coco, pgsql, saline, etc.).

## Solution

Introduce two new server-specific flags:
- `--server-image`: overrides the server container image name
- `--server-tag`: overrides the tag for the server container only; falls back to the global `--tag` if not provided

## Changes

- **`mgradm/shared/utils/flags.go`**: Added `ServerImageFlags` struct to `ServerFlags`; viper maps `--server-image` → `server.image` and `--server-tag` → `server.tag`, following the same pattern as other container flags
- **`mgradm/shared/utils/cmd_utils.go`**: Added `AddServerImageFlags()`, called from `AddServerFlags()` so the flags are available for install, upgrade and migration. Restored deleted comment explaining why some flags have no help group
- **`mgradm/cmd/install/install.go`**: Wired the override in `getFlagsUpdater` — if `Server.Image` or `Server.Tag` is set, it takes over `Image.Name`/`Image.Tag` respectively. Restored deleted comment on the podman alias args handling
- **`shared/podman/utils.go`**: Removed `Name`/`Tag` fields from `PodmanFlags` where they did not belong
- **Tests**: Added `ServerImageOnlyFlagsTestArgs` and `AssertServerImageFlags` covering the full override chain with distinct values, so the test genuinely fails if the wiring is broken. Updated `mgradm_install.go` and `server.go` test helpers accordingly

## Testing

- `--server-image` and `--server-tag` appear correctly in `mgradm install --help`
- Override chain verified: `--server-image custom/server` correctly populates `Image.Name`, independently of `--image`
- Backward compatibility confirmed: if `--server-tag` is not provided, the global `--tag` value is used
- Full test suite passes with zero failures


mgradm/cmd/install/install.go
```diff
@@ -22,6 +22,33 @@ type podmanInstallFlags struct {
...
+
+	// FIX #673: Server Tag Logic
+	flags.Podman.Name = v.GetString("server-image")
+
+	serverTag := v.GetString("server-tag")
+	if serverTag != "" {
+		flags.Podman.Tag = serverTag
+	} else {
+		flags.Podman.Tag = v.GetString("tag")
+	}
```
cbosdo: I don't think we need to shuffle around the image and tags here, rather fix where they are used…

mgradm/cmd/install/install.go
```diff
@@ -35,43 +62,22 @@ The command assumes podman is installed locally.
...
-			// If the alias "install podman" is used, "podman" will be the first arg.
-			// We remove it from the args slice so it isn't treated as the FQDN.
 			if len(args) > 0 && args[0] == "podman" {
 				copy(args, args[1:])
 				args = args[:len(args)-1]
 			}
 			return cobra.MaximumNArgs(1)(cmd, args)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// FIX #621: Pre-flight check
```
cbosdo: This comment doesn't bring any value.

mgradm/cmd/install/install.go
```diff
@@ -35,43 +62,22 @@ The command assumes podman is installed locally.
...
 			if _, err := exec.LookPath("podman"); err != nil {
 				return fmt.Errorf("podman is not installed. Please install podman before running this command")
 			}
 
 			var flags podmanInstallFlags
+			// We call the helper function here instead of defining the logic inline.
```
cbosdo: This comment doesn't bring value either.

cbosdo (changes requested):
I believe that the new image tags handling is brittle…

mihir-dixit2k27:
@cbosdo Thanks for the guidance! I have addressed all your feedback:

Refactor: I removed the manual flag shuffling in install.go and now rely on the mapstructure tags as you suggested.

Cleanup: I removed the verbose comments in cmd_utils.go and install.go.

Conflicts: I have resolved the conflicts with the recent refactor in main.

Regarding Squashing: Since I am working through the Web UI to ensure valid GPG signatures (due to local environment restrictions), I cannot easily squash these commits locally without breaking the signatures. Could you please use the "Squash and Merge" option when merging this PR?

Ready for final review!

cbosdo (changes requested):
This PR changes a crucial part of the logic (the image URLs computation) and there is no unit test added or adjusted. This looks really strange.

PR #734, opened 2026-03-04
feat: support multiple source files in mgrctl cp (Fixes #657)

### LINKED ISSUE #657: Idea: add wildcard support for mgrctl cp
Hi folks,
when using `mgrctl cp` only one file at the same time can be copied, e.g.:

```command
uyuni-server:~ # mgrctl cp a server:/root/a
10:01AM INF Starting mgrctl cp a server:/root/a
```

It would be handy having the possibility to copy multiple files at the same time. Currently, this is not possible as the `cp` subcommand only accepts 2 arguments:

```command
uyuni-server:~ # mgrctl cp a b server:/root/
Error: accepts 2 arg(s), received 3
```

Currently, a shell loop workaround is needed for this, e.g.:

```command
# for i in a b; do mgrctl cp $i server:/root/ ; done
10:05AM INF Starting mgrctl cp a server:/root/
10:05AM INF Starting mgrctl cp b server:/root/
```

YatesMold:
## What does this PR change?

Support multiple source files in `mgrctl cp` by accepting `MinimumNArgs(2)` instead of `ExactArgs(2)`, treating the last argument as the destination, and looping through all preceding arguments to copy each one.

Example: `mgrctl cp a b server:/root/` now works instead of requiring a shell loop.

## Test coverage
- No tests: already covered

## Links

Issue(s): #657

- [x] **DONE**

## Changelogs

- [x] No changelog needed


cbosdo (changes requested):
The PR looks quite good to me. I left some suggestions to make the help clearer, they may not be perfect, but it can't be left as is for sure.

The copyright year is not updated: add a copyright line with your name on the file you changed.

It would be probably nice to check that we don't have a mix of path starting with `server:` and not in the sources.

You also need to add a changelog file using `mkchlog`. see the [wiki contributing page](https://github.com/uyuni-project/uyuni/wiki/Contributing#changelog-rules-for-packages-in-git)

I would also need the PR description to be in the git commit message: it makes it easier to dig in code in the future.

YatesMold:
Hi @cbosdo, I've addressed all the review feedback: fixed the Use string and Long description, added a copyright line, added validation to reject mixed local/server sources, created a changelog entry, and expanded the commit message with the PR description. Please take another look when you get a chance

PR #755, opened 2026-03-31
Fix the macros in the spec file

cbosdo:
## What does this PR change?

Fix the suse_version macros and the macros in comments (bsc#1229105)

## Test coverage
- No tests: spec change

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/25070

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/uyuni-tools.spec
+++ b/uyuni-tools.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package uyuni-tools
 #
-# Copyright (c) 2025 SUSE LLC
+# Copyright (c) 2026 SUSE LLC
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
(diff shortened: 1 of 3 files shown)


uyuni-tools.spec
```diff
@@ -164,12 +164,12 @@ Summary:        Bash Completion for %{name_pxy}
 Group:          System/Shells
 Requires:       %{name_pxy} = %{version}
 BuildArch:      noarch
-%if 0%{?suse_version} >= 150000
+%if 0%{?suse_version} >= 1500
 Supplements:    (%{name_pxy} and bash-completion)
 %else
 Supplements:    bash-completion
```
aaannz: Maybe I misunderstood, but wasn't the bug also that we should not do this?

cbosdo: The issue is that those conditional supplements don't work for old distros. What I can do instead is to completely remove the Supplements in that case

aaannz: Yes, I think removing them is the solution.

cbosdo: Done

aaannz (approved):

PR #783, opened 2026-05-06
Fix Ubuntu builds before 26.04

cbosdo:
## What does this PR change?

BuildRequires issue on older Ubuntus

## Test coverage
- No tests: spec change

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

# Before you merge


uyuni-tools.spec
```diff
@@ -83,7 +83,11 @@ BuildRequires:  golang(API) >= 1.25
 
 %if 0%{?ubuntu}
 %define go_version      1.22
+%if 0%{?ubuntu} >= 2604
 BuildRequires:  golang >= %{go_version}
+%else
+BuildRequires:  golang-%{go_version}
+%endif
```
mcalmer: The problem is, that with the epoch handling this will never work.
What about this:

```suggestion
%define go_min_version      1.22
%if 0%{?ubuntu} >= 2404
# Will use the default golang of the OS which is higher than go_min_version
BuildRequires:  golang
%else
BuildRequires:  golang-%{go_min_version}
%endif
```

Ubuntu 24.04 and higher have regular golang packages which are good enough.
22.0 and lover we require the min version explicite.

PR #799, opened 2026-05-27
Improve readability of utils.YesNo for translators

marv7000:
## What does this PR change?

Improves the `utils.YesNo` function, allowing the strings to be translated easier.
Also added a check that makes the function return false when no response is given (as previously implied by `[y/N]`.

## Codespace

## Test coverage
- No tests: already covered

- [x] **DONE**

## Links

Issue(s): #

- [x] **DONE**

## Changelogs

- [x] No changelog needed

# Before you merge

### DIFF
--- a/shared/utils/utils.go
+++ b/shared/utils/utils.go
@@ -164,7 +164,8 @@ func AskIfMissing(value *string, prompt string, minValue int, maxValue int, chec
 func YesNo(question string) (bool, error) {
 	reader := bufio.NewReader(os.Stdin)
 	for {
-		fmt.Printf("%s [y/N]?", question)
+		// Translators: This is appended to the end of a question. Choices are short for `yes` and `no`.
+		fmt.Printf(L("%s [y/N]"), question)
 
 		response, err := reader.ReadString('\n')
 		if err != nil {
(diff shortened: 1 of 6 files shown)


shared/utils/utils.go
```diff
@@ -173,10 +174,14 @@ func YesNo(question string) (bool, error) {
...
+		// Empty responses imply `no`.
+		if len(response) == 0 {
+			return false, nil
+		}
+		if strings.HasPrefix(L("yes"), response) {
 			return true, nil
 		}
-		if strings.ToLower(response) == "n" || strings.ToLower(response) == "no" {
+		if strings.HasPrefix(L("no"), response) {
 			return false, nil
```
aaannz, line 185: This would accept only localized yes/no/y/n/... strings, right? How about also adding default english as well? For people with y/n muscle memory?

aaannz (approved):
LGTM, I have just small suggestion.

shared/utils/utils.go
marv7000: Yes, it would only accept the localized strings. I must say that so far I haven't seen any program do what you described. For example in German I only really see `j/N` accepted in programs that have proper translations.

marv7000: One potential issue I can think of would be that a language can have the same first letters for yes and no be the same, resulting in a `true` when the user meant to write the short form of no. I am not sure if this applies to any of our translations though

aaannz: You are right. That's probably reason why no one else does that.

PR #816, opened 2026-07-01
mgradm: Allow diskcheck env vars into containers (bsc#1270033)

mbussolotto:
## What does this PR change?

as title 

## Codespace

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/31099
5.1: https://github.com/SUSE/uyuni-tools/pull/258

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge

### DIFF
--- a/mgradm/cmd/restart/podman.go
+++ b/mgradm/cmd/restart/podman.go
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2025 SUSE LLC
+// SPDX-FileCopyrightText: 2026 SUSE LLC
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -20,6 +20,7 @@ func podmanRestart(
 	_ []string,
 ) error {
 	return utils.JoinErrors(
+		systemd.ReloadDaemon(false),
 		systemd.RestartService(podman.DBService),
 		systemd.RestartService(podman.ServerService),
 		systemd.RestartInstantiated(podman.ServerAttestationService),
--- a/uyuni-tools.changes.mbussolotto.diskcheck_fix
+++ b/uyuni-tools.changes.mbussolotto.diskcheck_fix
@@ -1,0 +1,1 @@
+- Reload systemd daemon before restarting services (bsc#1270033)


mgradm/shared/templates/serviceTemplate.go
```diff
@@ -43,6 +43,7 @@ ExecStart=/bin/sh -c '/usr/bin/podman run \
 	{{- range .Volumes }}
 	-v {{ .Name }}:{{ .MountPath }} \
 	{{- end }}
+	${PODMAN_EXTRA_ENV_VAR} \
```
cbosdo: What is the difference with `PODMAN_EXTRA_ARGS` ?

cbosdo (commented):
Please add description to the git commit messages… I wonder why we have two variables for the same thing…

mgradm/shared/templates/serviceTemplate.go
mbussolotto: I thought it was I separated PODMAN_EXTRA_ENV_VAR and PODMAN_EXTRA_ARGS for clarity, as I wanted to avoid overcrowding PODMAN_EXTRA_ARGS with too many different things (like env vars). This is just a first draft and my own opinion, though—I'm open to suggestions on what's best.

aaannz (commented):
Too bad that we did not name our env variables with well known prefix, like `UYUNI_`.
That way would could have plain `--env UYUNI_*` option and manage rest using service `Environment` values which could be easily added in drop-ins.

mbussolotto:
> Too bad that we did not name our env variables with well known prefix, like UYUNI_.

Do you think it might be too complicated to have something like this now? Not saying we should do it in this PR... just not sure how it might mess up the upgrade

aaannz:
I don't think it is complicated, but it may involve changing variables we have already documented and released.
But I didn't think too deeply about what everything need to be changed. 😄

mbussolotto:
@cbosdo @aaannz the daemon reload is required otherwise the changes on systemd file, required to read disckcheck env var, are not read

cbosdo (approved):

PR #834, opened 2026-08-12
Fix typos in shared SSL comments

RedZapdos123:
Fix three spelling mistakes in shared SSL and podman comments.

This is a comment only cleanup with no functional change.

Tested with `go test ./shared/podman ./shared/ssl ./shared/types`.

- [x] No changelog needed


cbosdo (approved):

cbosdo (changes requested):
You need to fix the copyright years as reported by the failed check

PR #850, opened 2026-09-09
Preserve the configured timezone during upgrade (bsc#1278137)

mbussolotto:
## What does this PR change?

The server timezone is only provided to the container through the TZ variable of the server.env environment file, which is applied on every container start.

On upgrade (and restore) GenerateUpgradeServerEnvironmentFile rewrote server.env from scratch with only the debug value, dropping the TZ line. The container then started without a timezone and fell back to UTC.

Preserve the timezone already configured for the server by reading it back from the environment file before regenerating it.

## Codespace

## Test coverage
- No tests

- [x] **DONE**

## Links

Issue(s): https://github.com/SUSE/spacewalk/issues/31958
5.2: https://github.com/SUSE/uyuni-tools/pull/282

- [x] **DONE**

## Changelogs

- [ ] No changelog needed

# Before you merge


cbosdo (approved):
LGTM, but aren't there other values we may want to keep in the future for upgrade? Why pruning that file if those values don't harm?

mbussolotto:
> LGTM, but aren't there other values we may want to keep in the future for upgrade? Why pruning that file if those values don't harm?

the other env var that might be used for the upgrade are:
- UYUNI_HOSTNAME
- MIRROR_PATH

the others are used just for installation and I think it's fine to don't have there

mbussolotto:
mmm but UYUNI_HOSTNAME should be change with rename command...so I think the only one missing is MIRROR_PATH

cbosdo:
> mmm but UYUNI_HOSTNAME should be change with rename command...so I think the only one missing is MIRROR_PATH

If they are needed, then I don't see them in the file generation here…

mbussolotto:
@cbosdo this should be better now

